PR #403, opened 2024-12-30
feat: close to tray, numerous tray improvements

braden-w:
This pull request introduces a significant overhaul of the tray service, including a new settings option to close to the tray and hide the window rather than fully closing the application.

## Known Issue on macOS
However, there is a known issue where, on macOS, Tauri's processes pause in the background due to the system's default policy that suspends background processes. This limitation impacts the ability to start and stop recordings on macOS specifically.

While this feature is expected to work on other platforms, it is advisable to test its functionality across different environments.

I dedicated extensive time to debugging this issue, but no fix has proven effective so far. As a workaround, it’s recommended to avoid close to the the tray on macOS for now. You can still keep the window behind other windows, just don't minimize it or hide it outright. Feel free to experiment and see what works!


mr-tchoo:
It would be great, if one could start Whispering minimized on startup, like LocalSend does it:
image

PR #633, opened 2025-08-12
Prefix app names with Epicenter.

Leftium:
Just prepending the brand name helps understand how everything is related.

- I wanted to color "Epicenter" black, but inline styles are stripped. (May be able to use image/SVG when we have more time...)
- So inserted `:` instead.
- Also changed second links because the name was already at the top.

If this looks OK, I will continue with the rest of the occurrences of app names in other README's.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,14 +56,14 @@ Epicenter is an ecosystem of open-source, local-first apps. Our eventual goal is
 <table>
   <tr>
     <td align="center" width="50%">
-      <h3>🎙️ <a href="https://github.com/epicenter-so/epicenter/tree/main/apps/whispering">Whispering</a></h3>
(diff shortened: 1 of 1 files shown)


braden-w:
CleanShot 2025-08-14 at 10 36 56@2x

Ahh unfortunately it seems like this really detracts from the readability of the README. Is it possible to condense this?

I'm not sure if we should go for a bullet-pointed list just because of how long it can be. I think it's probably worth including one or two extra sentences, but I prefer the old banner a little bit more just in terms of shape.

Leftium:
- I hid the bullet list behind a "More details" toggle.
- Renamed "Epicenter Shell" to "Epicenter Assistant"
  - I think most occurrences of "Epicenter Assistant" could be reduced to "Assistant"
  - Just add a note about the naming like for "Epicenter Whispering:"

image

braden-w:
I made a few minor edits, but otherwise, LGTM!

PR #636, opened 2025-08-13
fix(tooltip): coordinate tooltips globally and adjust delay timing

thisisharsh7:
This PR fixes two related tooltip behavior issues:

1. **Tooltip delay duration** — Reduced the delay so users see the tooltip almost instantly, without having to wait too long.
2. **Multiple tooltips visible** — When quickly hovering between different triggers, more than one tooltip could remain visible at the same time.

### Changes Made
- Set `delayDuration` to `300` and `skipDelayDuration` to `150` for faster tooltip display and instant hide behavior.
- Added logic to ensure only one tooltip can be open at any time when switching between triggers.

### Testing Steps
1. Hover over a tooltip — it should appear almost immediately.
2. Quickly hover between two or more tooltip triggers — only the most recent tooltip should be visible.

### Impact
Improves user experience by:
- Making tooltips appear without unnecessary waiting.
- Preventing overlapping tooltips when moving quickly between elements.

### DIFF
--- a/apps/whispering/src/lib/components/WhisperingTooltip.svelte
+++ b/apps/whispering/src/lib/components/WhisperingTooltip.svelte
@@ -25,18 +25,16 @@
 	{/if}
 {/snippet}
 
-<Tooltip.Provider>
-	<Tooltip.Root>
-		<Tooltip.Trigger {id}>
-			{#snippet child({ props: tooltipProps })}
-				{@render trigger?.({
-					tooltipProps: mergeProps(tooltipProps, restProps),
-					tooltip,
-				})}
-			{/snippet}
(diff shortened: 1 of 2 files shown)


braden-w:
Hey @thisisharsh7, thank you so much for this PR! 🎉

Your approach to coordinating tooltips globally was spot on. I've adapted your concept to the current codebase architecture in #1110 (the original `WhisperingTooltip.svelte` file no longer exists, so I couldn't merge directly).

You're included as a co-author on the commit, so you should see this contribution on your graph!

Thanks again for the inspiration and for improving the tooltip UX. Really appreciate it!

thisisharsh7:
thanks

PR #662, opened 2025-08-17
Convert prettier config files to JS for auto-format from VS Code.

Leftium:
Currently prettier auto-formatting does not work in VS Code. This is because VS Code's internal version of node does not yet support TS (config) files.

- This PR converts the TS config files to JS files with JSDocs comments to maintain benefits of types.
- Additionally the 'prettier-plugin-svelte' package had to be passed to prettier as a string instead of import object due to other problems in VS Code.
- Confirmed `bun run format` continues to work.

**Auto-complete thanks to types still works:**
image

### DIFF
--- a/apps/api/prettier.config.js
+++ b/apps/api/prettier.config.js
@@ -1,3 +1,4 @@
+/** @type {import('prettier').Config} */
 import { prettierConfig } from '@repo/config/prettier';
 
 export default prettierConfig;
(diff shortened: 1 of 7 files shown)


braden-w:
Hey @Leftium, LGTM! Thank you for this fix. 

Finally, when I save Svelte files they actually format properly. This was bugging me for a long time.

Thanks again!

PR #664, opened 2025-08-17
fix(Whispering): render Edit button at mobile breakpoints.

Leftium:
- fixes https://github.com/epicenter-so/epicenter/issues/658
- `<Modal>` expects both a `<Dialog.Root>` (desktop) and a `<Drawer.Root>` (mobile).
- `<Dialog>` renders on both.


Leftium:
**I investigated whether a Drawer would render better on mobile (phone screens) vs. dialog:**
1. Dialog does not render well on mobile (the dialog is too wide for the screen)
2. ~~HOWEVER: initial testing indicates the drawer would render about the same, since the parent page does not fit on mobile layouts and already overflows the page~~ the drawer may actually use the viewport width instead of the page width; except there is a gap on the right due to the scrollbar:
   

image

rupokghosh:
more context #658

Leftium:
Closing in favor of PR: https://github.com/epicenter-md/epicenter/pull/817

PR #703, opened 2025-08-21
fix(whispering): ensure notifications button has classes applied.

Leftium:
- Before this change, the `class` defined here (in the snippet) was being overwritten by the props being passed in.
- This ensures the button truly has fixed positioning as intended by the classes.
- **Note:** the `right-4` is also honored now, moving the button from the left to right corner. (The button will be covered by any toasts, making it inaccessible.)
- **Note:** the `z-50` is also honored, making the button appear above the modal overlay; however this is only visual. (Clicking the button will not activate the notifications log.)
- This PR fixes the same bug as: https://github.com/epicenter-so/epicenter/pull/673

# Before

**Notifications button is not hidden at xxs breakpoint.**
image

**Notifications button is not fixed; must scroll to bottom to find**
image

# After
**All classes honored, now: `fixed bottom-4 right-4 z-50 hidden xs:inline-flex`**
image

image

### DIFF
--- a/apps/whispering/src/lib/components/NotificationLog.svelte
+++ b/apps/whispering/src/lib/components/NotificationLog.svelte
@@ -45,12 +45,16 @@
 <Popover.Root bind:open={notificationLog.isOpen}>
 	<Popover.Trigger>
 		{#snippet child({ props })}
+			{@const { class: propsClass, ...restProps } = props}
 			<WhisperingButton
 				tooltipContent="Notification History"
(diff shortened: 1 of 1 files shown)


Leftium:
**Note:** Changes made in https://github.com/epicenter-so/epicenter/pull/612 may no longer be necessary.

(And perhaps https://github.com/epicenter-so/epicenter/pull/611?)

braden-w:
Great job @Leftium , LGTM! Thanks for going into the `bits-ui` internals.

Issue #710, opened 2025-08-24
Bug: Impossible to navigate between microphone screen and settings screen


zestyping:
Whispering has two completely different looks.  When I first opened it, I got a settings panel like this:

<img width="912" height="712" alt="Image" src="https://github.com/user-attachments/assets/ad44c30a-e9ae-4115-a515-884329be644f" />

When I opened it a second time, I got a microphone screen like this:

<img width="912" height="712" alt="Image" src="https://github.com/user-attachments/assets/23a8e72d-7a97-4bc3-acaa-efcd95cc9c83" />

The microphone screen has no navigation buttons, so it's impossible to get to the settings.  The settings screen has navigation buttons, but none of them go to the microphone screen and there is no close button, so it's impossible to get to the microphone screen.

Leftium:
It's good to know I'm not the only one! Thanks for your feedback. I have been advocating for a more consistent and user-friendly interface. (With labels next to the icons.)

**Navigation in the current app is unintuitive, but it *is* possible:**
- From the settings screen, clicking the "whispering" branding in the upper left corner takes you back to the microphone screen. (On hover, the tooltip says "Go home.")
- All the icons in the settings navbar exist on the microphone page; just in different locations:

<img width="802" height="602" alt="Image" src="https://github.com/user-attachments/assets/ffa5b10b-82a3-40ee-aa60-989b40d297b9" />

rupokghosh:
In my opinion, we should really think about the UX flow and have a consistent navigation system ( with centralized navigation components, philosophy, etc)

We should also have consistent icons, labels for the same routes/pages

thoughts? @Leftium

Leftium:
# Recommendations for Whispering (Navigation) UI
1. **Vertical sidebar navigation**
   - With labels next to icons.
   - May collapse to icons or even a single icon (but defaults to showing labels at wider breakpoints)
   - Example [shadcn-svelte block](https://www.shadcn-svelte.com/blocks/sidebar#sidebar-07) (sidebar that collapses to icons)
   - Clarity Design [Vertical Nav guide](https://clarity.design/documentation/vertical-nav).
   - See how Gmail web app does this (screen capture below)
2. **The Home page may become even cleaner by removing the current icon buttons in favor of the sidebar nav.**
   - The current icon buttons on the home screen may be left intact for convenience 
4. **Icons should be grouped and separated, possibly labeling the groups:**
   1. Navigation
   2. Config (Input, Model, Transformation selectors)
[...]

PR #713, opened 2025-08-26
Add active state highlighting to navigation items

rupokghosh:
**This PR marks the start of improving navigation overall in the codebase** 

- Right now, our navigation is kind of all over the place and has some issues with UX flow, see example ->  #710
- It's not that we have bad navigation, but it feels a bit unintuitive
- This PR highlights navigation items making it easier for the user to know where they currently are

> Before

Screenshot 2025-08-25 at 10 57 15 PM

> After

Screenshot 2025-08-25 at 10 56 08 PM

---

Screenshot 2025-08-25 at 10 56 53 PM


braden-w:
LGTM!

PR #717, opened 2025-08-26
Add CI workflow and improve formatting scripts

colecrouter:
Introduce a CI workflow to check the following:

- Format checks
- Lint checks
- Code checks

## Overview

- Added `.github/workflows/format.yml`
- Added `turbo.json` tasks:

| Name | Description | Goal |
|--------|--------------|------|
| `lint` | Runs global `biome lint` & `lint` package scripts | Run read-only linting |
| `check` | Runs `check` package scripts | Run code compile checks (`tsc`/`svelte-check`/whatever else) |
| `format` | Runs global `biome format --write` & `format` package scripts | Run code formatting |
| `format:check` | Runs `biome format` & `format:check` package scripts | Run read-only code formatting (meant specifically for CI) |

- Added/updated the above scripts in all packages
- Fixed easy TypeScript/linting errors
- Consolidated `.prettierignore`
- Added formatting for `.astro` files
- Added `.vscode` basic setup for DX

## TODO
- [X] Review warnings in each package; make sure there are no "bad formatting in /node_modules/whatever" or misconfigurations
- [ ] Get feedback to make sure all checks align with the team's needs
- [ ] Make sure consolidated `.prettierignore` doesn't wreck any projects
- [ ] Big `chore: format` commit
- [ ] Get configs for other editors!

### Fix Existing Errors 🤬

There are quite a few Biome/ESLint violations that will need to be reviewed before we can rely on this.

### ESLint & Biome Conflicts 😠

In our current set up, we are likely to run into duplicate warnings. I would suggest **disabling** ESLint for everything except `.svelte` files, which is in-line with the README for `@repo/packages/config`.

### Consolidate Configs 😴

We have an existing config package for Prettier and Eslint, but it may be worth looking into something consistent with TSConfig inheritence as well. Also, these are all running at the workspace level, but Biome is running at the package root. Not a blocking issue but should be investigated in the future.

### Delegating Dependencies 😴

If not already, we should develop a strategy to more consistently decide where packages should live. This aligns with the above issue.

### VSCode Config 😁

The current setup is quite simple:

#### Designate formatters for Specific File Types

No more—

> VSCode: "this fits on one line"
> Biome: "nuh uh"
> Prettier: "you guys are formatting on save?"

#### Organize Imports (Biome) on Save

If this is considered "too invasive", we can definitely leave this out. Personally, I would go all the way with "format with all plugins on save", as well as "remove unused imports" and "safe Biome fixes", and the whole kitchen sink; but I'm a chronic forgetter. Would love to hear what others think.


colecrouter:
## Update: Migrated ESLint & Prettier to Root

- ESLint and Prettier configs live at the root level
  - ESLint only handles `.svelte` files (as previously documented)
  - ESLint respects `.gitignore` (probably doesn't do anything, since it's only touching `.svelte`)
  - Prettier only touches formats BiomeJS doesn't touch (e.g. `.html`, `.svelte`)
- `lint`, `format`, `format:check`, and `check` scripts are optional in every package
  - `lint`, `format`, `format:check`, exist as manual overrides
  - `check` still required (see below)

**Next:** TypeScript configs ought to be consolidated more. We can either:
1. Share 1 config for all
  - Move `tsc` check to the root level
  - Make `check` an optional override, like the others
2. Inherit a root config
  - Keep package scripts as they are now

rupokghosh:
looking pretty good so far, should we be looking to break this down into multiple PRs for ease of reviewing?

colecrouter:
We could break it out. There isn't *all* that much; it's mostly just moving/copying/deleting/formatting files unfortunately, with only a handful of them being new. Whatever gets it merged sooner, though.

I'm happy leaving the other concerns for now, as my main goals are addressed anyway.

colecrouter:
@braden-w lmk what's blocking here. There aren't any logic *or* build-step changes. Just say the word and I'll rebase to a newer commit.

I've just noticed another PR addressing broken cross-package imports. We can't enforce that until we have the correct frameworks in place. Let's get this easy win 🏆

braden-w:
@colecrouter looks great! Can we rebase it whenever you have a chance? Otherwise I'll get on it!

colecrouter:
@braden-w 🙌 should be up-to-date now. ty!

braden-w:
@colecrouter Awesome, LGTM!

PR #726, opened 2025-08-29
Update accessibility settings modal flow and logic

rupokghosh:
### What problem are you trying to fix?

Initially I drafted out this PR to fix the accessibility page flow, we had no way of returning home or any kind of back button which felt extremely unintuitive to me, especially since the page is designed like a modal, I expected it to disappear if I click the background but that wasn't the case. 

Additionally, I found out that the notification block wasn't helpful at all since it wouldn't redirect you anywhere and just close the notification which users can find a bit annoying

Screenshot 2025-08-28 at 6 59 26 PM

### What did you do to fix the issue?

- I fixed the initial flow problem by adding a `Back to Home` button
- Added a notification message that tells the users actually what to do:

Screenshot 2025-08-28 at 7 00 43 PM

### DIFF
--- a/apps/whispering/src/routes/(config)/macos-enable-accessibility/+page.svelte
+++ b/apps/whispering/src/routes/(config)/macos-enable-accessibility/+page.svelte
@@ -1,37 +1,59 @@
 <script lang="ts">
 	import { Button } from '@repo/ui/button';
 	import * as Card from '@repo/ui/card';
-	import { SettingsIcon, CheckCircle2 } from '@lucide/svelte';
+	import { SettingsIcon, CheckCircle2, ArrowLeft } from '@lucide/svelte';
 	import * as services from '$lib/services';
 	import { toast } from 'svelte-sonner';
 	import { Command } from '@tauri-apps/plugin-shell';
 	import { createQuery } from '@tanstack/svelte-query';
 	import { rpc } from '$lib/query';
 	import type { PageData } from './$types';
+	import { goto } from '$app/navigation';
 
 	let { data }: { data: PageData } = $props();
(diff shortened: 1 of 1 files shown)


braden-w:
Hey @rupokghosh, thank you for this PR\! The macOS 14+ URL fix is spot on - that `x-apple.systemsettings` change was definitely needed.

I made one small adjustment to the error toast message. The new message suggested that "macOS 14+ restricts automatic opening of accessibility settings," but after digging into the `tauri-plugin-macos-permissions-api` implementation, I don't think that's accurate.

The plugin calls `requestAccessibilityPermission()` which should work fine on macOS 14+. If it fails, it's more likely due to other reasons (corrupt TCC database, app signing issues, user denying permission, etc.) rather than macOS 14+ specifically restricting this functionality.

Reverted back to the original error message: "Failed to open accessibility settings" - which is more accurate without making incorrect assumptions about macOS limitations.

The URL fix is solid though - really appreciate the contribution\! 🎉

braden-w:
I made a few additional improvements to this PR:

**UX Flow Improvements:**
- Removed the redundant inline link in the description that bypassed the proper "Request Permission" flow
- Now users are guided toward the programmatic permission request first, with manual settings as fallback

**Component Improvements:**  
- Replaced manual div styling with proper `Badge` component using `variant="success"`
- Switched from deprecated `CheckCircleIcon` to `CheckIcon`
- Better semantic HTML and design system integration

**Function Naming:**
- Renamed functions for clarity: `openAccessibilitySettings` → `openSystemSettings`, `requestAccessibilityPermission` → `requestPermissionOrShowGuidance`

The core URL fix remains solid - these are just polish improvements for better UX and code quality. Ready to merge! 🚀

PR #727, opened 2025-08-29
Refactor: Convert relative imports to absolute imports

rupokghosh:
Converted relative imports to absolute imports across many files using SvelteKit's $lib alias.

**Why this matters:**

- Easier navigation:` $lib/services/recorder` vs `../../../services/recorder` - immediately clear where code lives
- Safer refactoring: Moving files doesn't break import paths across the codebase
- Better for new developers: No need to mentally track complex directory hierarchies
- Reduced cognitive load: Focus on logic instead of calculating relative paths
- Improved IDE support: More reliable autocomplete and go-to-definition
- Future-proof: Scales better as codebase grows larger

We should also, in a future PR add this to our contributing guidelines so that we have consistent patterns throughout our entire codebase.


colecrouter:
I personally think this is super important. I also know that we can **enforce** this at the package level using turborepo, and inside our SvelteKit projects using [`eslint-plugin-boundaries`](https://github.com/javierbrea/eslint-plugin-boundaries).

If we can get #717 merged, I would love to FF this PR and have this solved permanently.

rupokghosh:
> I personally think this is super important. I also know that we can **enforce** this at the package level using turborepo, and inside our SvelteKit projects using [`eslint-plugin-boundaries`](https://github.com/javierbrea/eslint-plugin-boundaries).
> 
> If we can get #717 merged, I would love to FF this PR and have this solved permanently.

I back this too! Ty for the support @colecrouter

colecrouter:
Once we upgrade to Biome 2.x, we will be able to do something directly like this:

```jsonc
// biome.json
{
	"$schema": "https://biomejs.dev/schemas/2.2.4/schema.json",
	"overrides": [
		{
			"includes": [
				"apps/*/src/lib/**"
			],
			"linter": {
				"rules": {
					"style": {
						"noRestrictedImports": {
							"level": "warn",
							"options": {
								"patterns": [
									{
										"group": [
											"./*"
										],
										"message": "import module using `$lib/` alias"
									}
								]
							}
						}
					}
				}
			}
		},
		{
			"includes": [
				"apps/*/src/**"
			],
			"linter": {
				"rules": {
					"style": {
						"noRestrictedImports": {
							"level": "warn",
							"options": {
								"patterns": [
									{
										"group": [
											"**/lib/*"
										],
										"message": "import module using `$lib/` alias"
									}
								]
							}
						}
					}
				}
			}
		}
	]
}
```

I haven't crossed my Ts here, but I think something like this would achieve the desired outcome. I think 2.x also supports nested configs, so this could be placed elsewhere, such that it doesn't pollute the main config. Anyone else, feel free to chime in.

rupokghosh:
closing this since this seems like a stale PR now

PR #698, opened 2025-08-21
feat(linux-audio): add GStreamer backend for PipeWire compatibility

lockmeister:
## Summary

This PR fixes audio recording on Ubuntu 24.04+ with PipeWire by implementing a GStreamer-based audio backend that falls back to CPAL on other platforms.

Previously, users on modern Linux systems with PipeWire (like Ubuntu 24.04) experienced zero-length recordings due to CPAL compatibility issues. This implementation resolves that by:

- Adding a GStreamer-based audio recorder for Linux that works seamlessly with PipeWire
- Maintaining full backward compatibility with the existing CPAL implementation
- Providing automatic fallback from GStreamer to CPAL if GStreamer initialization fails

## Technical Details

**Key Changes:**
- Add GStreamer dependencies for Linux builds only (`gstreamer`, `gstreamer-audio`, `gstreamer-app`)
- Implement `GStreamerRecorder` with native PipeWire/PulseAudio support
- Create `AudioRecorderImpl` enum for clean backend abstraction
- Fix race condition in recording state management that was causing the zero-length issue
- Update device enumeration to work with both backends

**Architecture:**
The implementation uses a clean abstraction where:
1. On Linux: Try GStreamer first (for PipeWire compatibility)
2. On failure or other platforms: Fall back to CPAL
3. All existing APIs remain unchanged

**Testing:**
- Tested on Ubuntu 24.04 with PipeWire 
- Verified working audio recording with both `pulse` and `pipewire` device selections
- Confirmed CPAL fallback works on systems without GStreamer
- No regressions on existing functionality

## Test Plan

- [x] Test on Ubuntu 24.04 with PipeWire - audio recording works
- [x] Test device enumeration shows available audio sources  
- [x] Test recording produces non-zero length files with actual audio data
- [x] Verify clean fallback to CPAL when GStreamer unavailable
- [x] Confirm no regressions on existing platforms

Fixes #697

Co-authored-by: lockmeister <lockmeister@users.noreply.github.com>

PR #729, opened 2025-08-29
fix: remove unnecessary security-related changes from GStreamer PR

braden-w:
## Summary

This PR removes unnecessary and potentially harmful changes from #698 while keeping the excellent GStreamer implementation intact.

## Context

PR #698 introduces a great GStreamer-based audio backend that fixes PipeWire compatibility issues on Ubuntu 24.04+. However, it includes some unrelated changes that appear to be debugging artifacts and introduce security concerns.

### 1. Reverted CPAL to stable version (0.16.0)
- The PR switched from stable `cpal = "0.16.0"` to git master branch
- This is unnecessary since the GStreamer implementation already solves the PipeWire issue
- Git dependencies can break with any commit and should be avoided in production

### 2. Removed browser-specific arguments
The following WebView arguments were removed as they're unrelated to native audio recording:
- `--enable-media-stream`: For WebRTC/browser media APIs
- `--use-fake-ui-for-media-stream`: Auto-approves browser permissions (security risk)
- `--disable-web-security`: Disables CORS and web security (major security risk)
- `--allow-running-insecure-content`: Allows HTTP on HTTPS (security risk)

### 3. Removed CSP modification flag
- Removed `dangerousDisableAssetCspModification: true`
- This flag disables Content Security Policy modifications and is unrelated to audio recording

## What's Kept

All the excellent GStreamer implementation remains intact:
- Full GStreamer recorder implementation
- Smart fallback from GStreamer to CPAL
- Linux-specific conditional compilation
- Proper PipeWire/PulseAudio support

## Testing

The GStreamer audio recording should work exactly as before, but without the security vulnerabilities introduced by the browser arguments.

## Discussion

@lockmeister - Great work on the GStreamer implementation\! I've created this PR to isolate just the security-related cleanup. If these changes were intentional for a specific use case, we can discuss alternatives that don't compromise security.

Builds on #698


braden-w:
Closing this pull request in favor of #820. Hopefully that fixes the same issue!

Your cleanup work on the GStreamer implementation was excellent, but #820 takes an even simpler approach by addressing the root cause directly with sample format filtering. This avoids the need for GStreamer dependencies entirely while solving the same PipeWire compatibility issue on Ubuntu 24.04+.

PR #734, opened 2025-08-30
fix(EditRecordingModal): prevent audio rerendered

thisisharsh7:
This fixes the audio player rerendering
  every time users type in the edit
  recording modal input fields. Previously,
   typing in the title, subtitle, or other
  text fields would cause the audio player
  to reload and reset its playback
  position, creating a poor user
  experience.

  The issue occurred because the audio
  player's `src` was derived from
  `workingCopy.blob`, but `workingCopy`
  gets updated on every keystroke. Since
  the audio blob itself never changes
  during editing (users are only modifying
  metadata), the audio player was
  unnecessarily rerendering.

before: 

https://github.com/user-attachments/assets/2ef9e4b7-7e2a-4df2-9550-50a765106207

after: 

https://github.com/user-attachments/assets/cd0cdcb9-9da5-45d1-89eb-5d6fc8492e61

### DIFF
--- a/apps/whispering/src/routes/(config)/recordings/row-actions/EditRecordingModal.svelte
+++ b/apps/whispering/src/routes/(config)/recordings/row-actions/EditRecordingModal.svelte
@@ -70,8 +70,8 @@
 	const blobUrlManager = createBlobUrlManager();
 
 	const blobUrl = $derived.by(() => {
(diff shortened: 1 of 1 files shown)


braden-w:
LGTM! Thank you :)

PR #736, opened 2025-08-31
feat(completion): add OpenRouter support with OpenAI SDK integration

braden-w:
Hey @vishesh, thank you for adding OpenRouter support\! This is a great addition that gives users more flexibility in choosing their completion providers.

I made some improvements to the implementation that I wanted to run by you:

### Migrated from fetch to OpenAI SDK
Since OpenRouter is 100% compatible with the OpenAI API format, I refactored the implementation to use the OpenAI SDK instead of raw fetch calls. This gives us:

- **Full TypeScript type safety** - No more manual type definitions for API responses
- **Automatic retry logic** - Built-in exponential backoff for transient failures  
- **Better error handling** - Structured error types from the SDK
- **Consistency** - Matches our existing OpenAI implementation pattern

### Implementation Details
The key change is configuring the OpenAI client with OpenRouter's base URL:
```typescript
const client = new OpenAI({ 
  apiKey,
  baseURL: 'https://openrouter.ai/api/v1',
  dangerouslyAllowBrowser: true,
  defaultHeaders: {
    'HTTP-Referer': 'https://whispering.epicenter.so',
    'X-Title': 'Whispering',
  }
});
```

### Enhanced Error Handling
Added comprehensive error handling for OpenRouter-specific cases:
- **402 Payment Required** - When users run out of credits
- **502 Bad Gateway** - When model providers are unavailable (OpenRouter handles fallbacks)
- All standard HTTP error codes with user-friendly messages

## Benefits
- Less code to maintain (removed manual fetch logic)
- Type-safe API interactions
- Future-ready for streaming support if needed
- Same interface so no breaking changes

The core functionality you built remains the same - this just makes it more robust and maintainable. Let me know if you have any questions about the changes\!

Thanks again for the contribution\! 🎉


vishesh-sachan:
hey @braden-w, That's great to hear! I'm glad you think the changes are a solid improvement. Using the OpenAI SDK for OpenRouter makes the integration much more reliable and easier to maintain with type safety and automatic retries. Thanks again for the feedback.

PR #774, opened 2025-09-08
Add simple error page for 404 and other errors

rupokghosh:
I realized that we didn't have a dedicated error page so therefore I added a very simple version of the page. 

> Before 

Screenshot 2025-09-07 at 7 03 47 PM

> After
Screenshot 2025-09-07 at 7 03 23 PM


braden-w:
LGTM! Made a small fix to use the new Svelte 5 Runes instead of stores.

PR #787, opened 2025-09-15
Fix(docs): point docs to our docs and server

rupokghosh:
Previously, we were pointing to Astro documentation in both of the `docs` and `discord` links in the PR, fixed it by pointing it to the correct documentation.


braden-w:
LGTM!

PR #799, opened 2025-09-16
feat(landing-page): add Try Now buttons to landing page tool cards

rupokghosh:
Added a couple of new buttons to not only point to source code but also how the user can quickly access the application. 

- For whispering - I am pointing them to the releases page. 
- For epicenter.sh - I am pointing them to the home page of sh

Screenshot 2025-09-16 at 3 39 14 PM


braden-w (approved):
LGTM - great feature with improved data structure

PR #817, opened 2025-09-22
Fix: modal trigger not rendered on mobile breakpoints.

Leftium:
Updated vaul-svelte: the older version must not be compatible with the other UI packages/code.

Fixes: https://github.com/epicenter-md/epicenter/issues/658


braden-w:
LGTM!

PR #844, opened 2025-09-29
feat: auto enter on over

thanhdatle:
# Add automatic Enter key press on "Over" voice command

  This feature enables hands-free message submission by automatically pressing Enter when transcribed audio ends with "Over".

  ## What it does

  When recording audio that ends with "Over" (case-insensitive, with optional punctuation), the system will:
  1. Remove "Over" from the transcribed text before delivery
  2. Deliver the cleaned text normally (clipboard/cursor writing)
  3. Automatically press Enter to submit the message

  ## Use case

  Perfect for voice-to-chat workflows with Claude or other interfaces where you want to record a message, have it transcribed and automatically submitted without
  manual interaction.

  **Example:**
  - Say: "Hello Claude, can you help me debug this code? Over."
  - Result: "Hello Claude, can you help me debug this code?" → *Enter pressed automatically*

  ## Implementation

  **Backend (Rust):**
  - Added `press_enter()` function using existing `enigo` library with cross-platform key codes
  - Registered new Tauri command handler

  **Frontend (TypeScript):**
  - Extended text service interface with `pressEnter()` method
  - Added detection utilities in `transcription-triggers.ts`:
    - `shouldTriggerEnterOnOver()` - detects "Over" at text end
    - `removeOverFromEnd()` - cleanly strips "Over" and punctuation
  - Integrated into delivery pipeline to process text before clipboard/cursor operations

  **Flow:**
  1. Transcription completes with "Over" detection
  2. Text gets cleaned (Over removed) before any delivery operations
  3. Normal delivery proceeds with cleaned text (clipboard, notifications, etc.)
  4. Enter key automatically pressed after 200ms delay

### DIFF
--- a/apps/whispering/src-tauri/src/lib.rs
+++ b/apps/whispering/src-tauri/src/lib.rs
@@ -70,6 +70,7 @@ pub async fn run() {
     // Register command handlers (same for all platforms now)
     let builder = builder.invoke_handler(tauri::generate_handler![
         write_text,
+        press_enter,
         // Audio recorder commands
         get_current_recording_id,
         enumerate_recording_devices,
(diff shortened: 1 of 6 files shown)


rupokghosh:
What happens when I am transcribing something and it actually has the word over? does it stop and press enter? That would not be an ideal behaviour.

thanhdatle:
> What happens when I am transcribing something and it actually has the word over? does it stop and press enter? That would not be an ideal behaviour.

Thanks for asking the question.I'm using whispering to answer the question. I say over, but it doesn't trigger the Enter key.Only when I say over at the end.like this.
And it will hit the enter key.

PR #854, opened 2025-10-03
Fix CI setup for formatting

colecrouter:
@Leftium found out that the current `format` script is running both BiomeJS  & Prettier on Svelte/Astro files ([see this link from Biome docs](https://v1.biomejs.dev/internals/language-support/)). This is fine but prevents `format:check` from succeeding.

This commit adds an override to the BiomeJS config, which skips formatting for:
- Semi-supported file formats
- Files we use other formatters on


colecrouter:
For clarity:
image

colecrouter (commented):
I just rebased this branch; it now matches the use of Biome v2 on `main`.

PR #863, opened 2025-10-06
fix(LocalModelDownloadCard): prevent showing models as active when not downloaded

braden-w:
This fixes an issue where local model download cards could show conflicting states: displaying the "Active" badge while also showing a "Download" button, or showing pre-built models as active when they weren't actually downloaded.

The root cause was insufficient validation in the `refreshStatus()` function. The original implementation only checked if the settings path matched the expected model path, but didn't verify that the settings path was non-empty or that the file actually existed.

This created scenarios where:
- A user manually selected a model, then deleted the file, but the UI still showed it as active
- Settings pointed to a different model path, but pre-built models incorrectly appeared active
- The "Active" badge appeared on models that were selected in settings but never downloaded

The fix adds three validation checks before marking a model as active:

1. Settings path must not be empty
2. Settings path must exactly match this model's expected download path
3. File must actually exist at that path (verified via `isModelValid()`)

After this change, model states display correctly:
- "Active" badge + "Activated" button: Model is downloaded AND currently selected
- "Downloaded" badge + "Activate" button: Model is downloaded but not selected  
- No badge + "Download" button: Model is not downloaded

The additional existence check ensures the UI stays in sync with the actual filesystem state, preventing stale settings from causing incorrect UI states.

### DIFF
--- a/apps/whispering/src/lib/components/settings/LocalModelDownloadCard.svelte
+++ b/apps/whispering/src/lib/components/settings/LocalModelDownloadCard.svelte
@@ -111,15 +111,26 @@
 				const path = await ensureModelDestinationPath();
 				const isValid = await isModelValid(path);
 
+				// If the model doesn't exist at the expected path, it's not downloaded
 				if (!isValid) {
 					modelState = { type: 'not-downloaded' };
 					return;
 				}
 
-				// Check if this model is active in settings
+				// Model exists at expected path, now check if it's the currently selected model
 				const settingsKey = `transcription.${model.engine}.modelPath` as const;
 				const currentPath = settings.value[settingsKey];
-				const isActive = currentPath === path;
+				
+				// Only consider active if:
+				// 1. Settings path is not empty
+				// 2. Settings path exactly matches this model's expected path
+				// 3. The file at the settings path actually exists (additional safety check)
+				let isActive = false;
+				if (currentPath && currentPath === path) {
+					// Double-check that the file at the settings path actually exists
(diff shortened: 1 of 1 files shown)


braden-w:
I took a careful look at this PR and traced through the logic. I don't think this actually fixes a bug that exists in the codebase.

## Why the described bugs can't actually occur:

**Scenario 1: "Model deleted but still shows active"**
- The existing code already has an early `isValid` check at line 112
- If the file doesn't exist, we return early with `modelState = 'not-downloaded'`
- We never reach the `isActive` comparison

**Scenario 2: "Pre-built models showing as active when custom model selected"**
- Original code: `isActive = currentPath === path`
- If `currentPath = "/custom/path"` and `path = "/app-data-dir/model-tiny.bin"`
- The comparison returns `false` correctly

**Scenario 3: "Conflicting states (Active badge + Download button)"**
- This can't happen because the early `isValid` check prevents reaching the active state logic

## Why the new checks are redundant:

1. **Empty string check** (`currentPath &&`): In JavaScript, `"" === path` is always `false` (since `path` is never empty). This doesn't change behavior.

2. **Double file existence check**: We already verified `isModelValid(path)` earlier. If `currentPath === path`, we've already confirmed the file exists. Checking again with `isModelValid(currentPath)` is redundant.

This is defensive programming that makes the code more explicit, but it doesn't actually fix a bug. The original logic already handles all these cases correctly.

Closing this PR since it doesn't address an actual issue.

PR #885, opened 2025-10-14
feat(transcription): implement persistent model manager for Parakeet

braden-w:
This implements a persistent model manager that keeps the Parakeet transcription model loaded in memory between transcription calls, resulting in significant performance improvements for typical usage patterns where users perform multiple transcriptions in a session.

The current implementation loads the Parakeet model from disk for every transcription request, taking approximately 1.5 seconds for model loading plus 0.3 seconds for the actual transcription. This means every transcription takes about 1.8 seconds total, with most of that time spent on redundant model loading.

The persistent model manager solves this by loading the model once and reusing it for subsequent transcriptions. After the initial load, transcriptions complete in just 0.3 seconds, providing a 5x speedup. The model automatically unloads after 5 minutes of inactivity to prevent unnecessary memory usage.

This change uses Tauri's state management to maintain a single ModelManager instance across the application lifecycle. The ModelManager uses Arc and Mutex for thread-safe access to the loaded model, ensuring that concurrent transcription requests are handled safely. The implementation intelligently handles model switching by unloading the current model when a different model path is requested.

From a memory perspective, the loaded model adds approximately 500MB of RAM usage. The 5-minute idle timeout provides a good balance between keeping the model available for active users while freeing memory when the feature is not in use. Users performing multiple transcriptions within a few minutes will see dramatic performance improvements without any manual configuration needed.

### DIFF
--- a/apps/whispering/src-tauri/src/lib.rs
+++ b/apps/whispering/src-tauri/src/lib.rs
@@ -9,7 +9,7 @@ use recorder::commands::{
 };
 
 pub mod transcription;
-use transcription::{transcribe_audio_whisper, transcribe_audio_parakeet};
+use transcription::{transcribe_audio_whisper, transcribe_audio_parakeet, ModelManager};
 
 pub mod windows_path;
 use windows_path::fix_windows_path;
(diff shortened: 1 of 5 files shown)


ljbred08:
Thank you so much! This is a massive improvement and makes Parakeet actually usable (I wondered why I wasn't getting the speeds I've seen in my own parakeet tests!)

PR #896, opened 2025-10-17
fix(ffmpeg): prevent race condition on Windows with file verification

ghul0:
This adds file verification after writing temporary files to force Windows OS buffer flushing, preventing race conditions where FFmpeg attempts to read files before they're fully written to disk.


braden-w:
LGTM! Thank you :)

Issue #815, opened 2025-09-21
Command prompt flashes on every recording with Parakeet


rho-cassiopeiae:
### Bug:
Every time I record using Parakeet, command prompt flashes for an instant . Transcription works fine, but the flickering is annoying.

### Platform

Windows

### Affected Component

Transcription

### Whispering Version

v7.5.0

### Severity

Minor (cosmetic issue or minor inconvenience)

### Checklist

- [x] I have searched existing issues to ensure this hasn't been reported
- [x] I have included all relevant information above

VijayPL:
Thank you for the report. I'll let the dev team know about this.

jakub-wojciech:
For me, the window I was in loses focus. I need to click on it again for the text to be pasted. This makes "parakeet" not usable.
I am on Windows 11.

Leftium:
There is another [report](https://discord.com/channels/1391098486178582549/1404933147812040796/1419985358728200255) (from Discord) that indicates this may be related to when Whispering calls an external binary (like FFmpeg) via the shell.

**Screen shot from user:**

<img width="1225" height="607" alt="Image" src="https://github.com/user-attachments/assets/d97ab79b-6520-4948-ad07-b94f75c98441" />

PR #909, opened 2025-10-22
fix(windows,transcription): prevent FFmpeg console window flash

brummelte:
## Summary

  Prevents the brief Windows console window flash when local transcription engines run FFmpeg during audio
  conversion.

  ## Context

  Local engines (whispercpp and parakeet) share a Rust converter that uses FFmpeg for compressed inputs. On Windows, that FFmpeg process was spawned without console-hiding flags, causing a momentary console window.

  ## What Changed

  - Apply Windows-only CREATE_NO_WINDOW when spawning FFmpeg.
  - Refactor to use a mutable Command before .output() so the flag can be set.

  ## Scope

  - Affects only local engines: whispercpp, parakeet (both use convert_audio_for_whisper).
  - Platform-specific: Windows only.
  - No API/signature changes; behavior is unchanged aside from hiding the flash.

## References

References #815
References #899

@braden-w

### DIFF
--- a/apps/whispering/src-tauri/src/transcription/mod.rs
+++ b/apps/whispering/src-tauri/src/transcription/mod.rs
@@ -4,6 +4,8 @@ mod model_manager;
 use error::TranscriptionError;
 pub use model_manager::ModelManager;
 use std::path::PathBuf;
+#[cfg(target_os = "windows")]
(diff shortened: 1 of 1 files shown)


braden-w:
Thank you! LGTM :)

PR #979, opened 2025-11-11
feat: add window state persistence

vishesh-sachan:
## Summary
Adds automatic window state persistence to remember window size and position across app restarts.

## Type of Change

- [x] `feat`: New feature

## Related Issue

Closes #878 

## Changes Made

- Added `tauri-plugin-window-state` v2.4.0 dependency
- Registered window-state plugin in Rust backend
- Window size, position, and state (maximized/minimized) now persist automatically

The plugin uses Tauri v2's default permission system - all window-state operations are enabled by default with no additional configuration needed. State is saved automatically on window changes and restored on app launch.

## Testing

1. Resize/move the window
2. Close the app
3. Reopen - window should restore to previous size and position

### Desktop App Testing
- [x] Tested on macOS
- [ ] Tested on Windows
- [ ] Tested on Linux

### DIFF
--- a/apps/whispering/src-tauri/Cargo.lock
+++ b/apps/whispering/src-tauri/Cargo.lock
@@ -5473,6 +5473,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "tauri-plugin-window-state"
+version = "2.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


Leftium:
I often intentionally close Whispering and re-open it to get it back to the the "default" size. So a way to do this would be appreciated. Perhaps another button like the "minimize" button.

PR #984, opened 2025-11-13
feat(whispering:transformations): add custom OpenAI-compatible provider

thurstonsand:
Adds support for custom OpenAI-compatible endpoints in transformation steps. Users can now point to Ollama, LM Studio, vLLM, reverse proxies, or any `/v1` compatible endpoint without waiting for dedicated integrations.

While implementing this, I consolidated the OpenAI and OpenRouter services which were basically the same thing except with different base urls. They now use a shared `openai-compatible.ts` factory.

Should address #692, #837, #821

### New Custom Provider

- Configure any OpenAI-compatible endpoint with optional API key
- Global settings + per-step base URL overrides

### Refactored Services

- `openai-compatible.ts`: Factory for OpenAI-compatible services with configurable error messages, headers, and status overrides
- `openai.ts` and `openrouter.ts`: Refactored to use the factory
- `custom.ts`: New service for custom endpoints

### UI

- `CustomEndpointInput.svelte` for Settings → API Keys
- Transformation editor integration with per-step configuration

## Configuration

Set default endpoint in Settings → API Keys → Custom API Endpoint. Override per-step if needed.

Example: `http://localhost:11434/v1` with model `gemma3:latest` or `mistral:latest`

## Screenshots

Screenshots taken in the transformations modal, after configuring each provider, and successfully transforming some text given a basic prompt.

### openai still works
Screenshot 2025-11-12 at 21 22 56

### openrouter still works
Screenshot 2025-11-12 at 21 22 06

### new custom endpoint works
Screenshot 2025-11-12 at 21 19 31


braden-w:
Hey @thurstonsand, thank you for this excellent PR! The custom endpoint feature is working great.

I did some follow-up refactoring in #998 to consolidate the factory pattern across all OpenAI-compatible services. The refactoring eliminates the wrapper pattern and moves provider-specific logic (like baseUrl resolution and validation) into factory config functions. This makes it much easier to add new OpenAI-compatible providers in the future.

Would love your thoughts on the approach if you get a chance to review!

PR #791, opened 2025-09-15
feat: migrate to transcribe-rs for Local Transcription

braden-w:
# Migrate to transcribe-rs for Local Transcription

## Summary
This PR migrates our local transcription from using whisper-rs directly to using the more flexible transcribe-rs library, fixing GPU issues and simplifying our architecture.

## Changes
- **Replaced whisper-rs with transcribe-rs** - Now using @cjpais's unified transcription library
- **Simplified dependencies** - Removed platform-specific Metal/Vulkan feature flags
- **Improved UI feedback** - Temperature and prompt fields now clearly indicate when they're not supported for local models
- **Added acknowledgments** - Properly credited the amazing open-source projects powering our local transcription

## Technical Details
- transcribe-rs provides a unified API over multiple speech-to-text engines (Whisper and Parakeet)
- Temperature and initial prompt parameters are not available in transcribe-rs (these are Whisper-specific features that don't translate to other engines like Parakeet)
- UI gracefully handles this by disabling these fields for local models while keeping them enabled for cloud providers

## Benefits
- Cleaner, more maintainable codebase
- Future-ready for adding Parakeet or other local engines
- Automatic GPU acceleration handled by the library
- Consistent API across different transcription engines

## Acknowledgments
Special thanks to @cjpais for creating transcribe-rs and Handy, which made this migration possible and solved our GPU issues elegantly.

Fixes GPU-related issues with local transcription.

PR #1006, opened 2025-11-27
fix(windows): bundle vulkan-1.dll to fix startup crash

braden-w:
Windows users without the Vulkan runtime installed were experiencing "vulkan-1.dll not found" errors that prevented the app from starting. This affected Windows 10 systems, machines without dedicated GPUs, virtual machines, and clean Windows installations.

The root cause: our migration to transcribe-rs (PR #791) introduced a hard dependency on Vulkan through whisper-rs. Windows checks all DLLs in the import table at process startup, so the app fails immediately before any code runs.

The fix bundles vulkan-1.dll from the Vulkan SDK during CI builds. Since Tauri's `bundle.resources` places files in a `resources/` subdirectory rather than next to the executable, I added NSIS installer hooks that move the DLL to the correct location after installation. Once the DLL is present, whisper.cpp's GGML backend gracefully falls back to CPU transcription on systems without GPU/Vulkan drivers.

Note: This fix works for NSIS installers (the default). MSI installers don't support the same hook mechanism, but NSIS is the primary distribution method.

Fixes #829, fixes #840

### DIFF
--- a/.github/workflows/pr-preview-builds.yml
+++ b/.github/workflows/pr-preview-builds.yml
@@ -94,6 +94,18 @@ jobs:
           version: 1.4.309.0
           cache: true
 
+      # Bundle Vulkan DLL for Windows users without Vulkan runtime
+      - name: Bundle Vulkan DLL (Windows)
+        if: matrix.platform == 'windows-latest'
+        shell: pwsh
+        run: |
+          $vulkanDll = "$env:VULKAN_SDK\Bin\vulkan-1.dll"
+          $destination = "apps/whispering/src-tauri/vulkan-1.dll"
+          Write-Host "Copying Vulkan DLL from: $vulkanDll"
+          Write-Host "Copying Vulkan DLL to: $destination"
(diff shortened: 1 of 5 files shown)


braden-w:
Closing this PR as superseded by #1181 and #1182.

The underlying problem (vulkan-1.dll dependency on Windows) has been addressed by a different approach: disabling whisper-cpp on Windows entirely. This was necessary because of fundamental MSVC runtime conflicts between `whisper-rs-sys` (MD_DynamicRelease) and `tokenizers/esaxx-rs` (MT_StaticRelease).

**Current Windows local transcription options:**
- ✅ Parakeet (available)
- ❌ Whisper.cpp (disabled due to CRT conflicts)
- ❌ Moonshine (disabled due to CRT conflicts)

This is a temporary limitation while we work on resolving the cross-compilation issues with these local transcription libraries. See #1181 and #1182 for full context.

Fixes #829, Fixes #840

PR #1042, opened 2025-11-29
refactor: rename transcribedText field to transcript

braden-w:
This renames the Recording type's `transcribedText` field to `transcript` throughout the data layer. The UI layer (labels, component names, dialog titles) had already been updated, but the underlying data field still used the old name. This creates consistency between the user-facing terminology and the codebase internals.

The change touches 12 files across type definitions, serialization logic, query layer mutations, and component property accesses. Because this is a breaking change to the Recording type, all updates needed to happen atomically.

Key areas updated:
- Type definition in `models/recordings.ts`
- Markdown serialization/deserialization in `file-system.ts` (frontmatter + body pattern)
- Query layer files (`transcription.ts`, `actions.ts`, `transformer.ts`)
- All Svelte components that access the field
- View transition `propertyName` identifiers
- Table column `accessorKey` configurations

Note: README documentation files contain example code snippets that still reference `transcribedText`. These are illustrative examples and updating them is optional since they don't affect runtime behavior.

### DIFF
--- a/apps/whispering/src/lib/components/MigrationDialog.svelte
+++ b/apps/whispering/src/lib/components/MigrationDialog.svelte
@@ -79,7 +79,7 @@
 				'This is a medium-length recording with a bit more content to transcribe and process.',
 				`This is a longer recording transcript. It contains multiple sentences and paragraphs of content. ${Array(10).fill('Lorem ipsum dolor sit amet, consectetur adipiscing elit.').join(' ')}`,
 			];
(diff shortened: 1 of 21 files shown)


braden-w:
Closing this as absorbed by newer work.

Live Whispering schema stores `transcript` already. Remaining `transcribedText` hits in current source are local variable names around transcription results, not persisted row fields. No PR work is left here.

Reference: `apps/whispering/src/lib/workspace/definition.ts`.

PR #1046, opened 2025-11-29
feat: add option to simulate Enter keystroke after transcription

braden-w:
This adds a new setting that allows users to automatically press Enter after transcription or transformation text is pasted at the cursor. This is useful for chat applications like Slack, Discord, or iMessage where pressing Enter submits the message.

The implementation uses the `enigo` crate (already used for paste simulation) with `Key::Return` and `Direction::Click`, which DeepWiki confirmed is the idiomatic approach. The feature is desktop-only since browsers cannot simulate keystrokes for security reasons.

The setting only appears in the UI when the parent "paste at cursor" setting is enabled, and is gated behind `window.__TAURI_INTERNALS__` so web users don't see an option they can't use.

Closes #720


piknockyou:
Many thanks!

PR #1234, opened 2026-01-10
Fix: Debounce system prompt saves to prevent toast spam

### LINKED ISSUE #1214: System prompt success notification on each keystroke
### Bug:

When changing the system prompt, every keystroke gets saved and you see a new toast notification for it.

### Expected behavior:

I expect to see a notification when I finish typing the entire thing rather than at each keystroke.

### Steps to reproduce:
1. Go to the settings page.
2. Go to the system prompt section.
3. Start typing anything.

### Additional context:

Image

### Platform

Mac OS

### Affected Component

UI-App (desktop interface)

### Whispering Version

v7.11.0

### Severity

Minor (cosmetic issue or minor inconvenience)

### Discord Link

_No response_

### Checklist

- [x] I have searched existing issues to ensure this hasn't been reported
- [x] I have included all relevant information above

vishesh-sachan:
### Problem
Editing the System Prompt in transcription settings triggered a save on **every keystroke**, resulting in constant toast notifications and a poor user experience.

### Solution
Implemented **debouncing** for the save action:

- Added a reusable debounce utility via `@epicenter/svelte-utils`
  - Svelte 5–compatible
  - Simple API + cleanup method for proper teardown
- Applied debounce to the System Prompt field
  - Local state manages textarea value
  - Save is delayed **500ms after the last keystroke**
  - Cleanup is handled on component destroy

### Result
Users can type freely without toast spam. A single toast is shown only after typing stops for ~500ms.

---

## Type of Change

- [x] `fix`: Bug fix

---

## Related Issue

Closes #1214 

---

## Testing Instructions

1. Navigate to **Settings → Transcription**
2. Edit the **System Prompt** field
3. Confirm that:
   - Only one toast appears after you stop typing
   - No toasts are shown on every keystroke

---

## Screenshots / Recordings

https://github.com/user-attachments/assets/78178bde-6831-4d88-a3dd-5572947d96c9

### DIFF
--- a/apps/whispering/src/routes/(app)/(config)/settings/transcription/+page.svelte
+++ b/apps/whispering/src/routes/(app)/(config)/settings/transcription/+page.svelte
@@ -34,6 +34,8 @@
 	import * as Select from '@epicenter/ui/select';
 	import { Textarea } from '@epicenter/ui/textarea';
 	import { hasNavigatorLocalTranscriptionIssue } from '$routes/(app)/_layout-utils/check-ffmpeg';
+	import { debounce } from '@epicenter/svelte-utils';
+	import { onDestroy } from 'svelte';
 
 	const { data } = $props();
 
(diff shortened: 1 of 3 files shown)


Leftium:
Is there a reason for the system prompt to be saved every keystroke?
- I think it would be better UX if the dialog only saved a single time: after the user finished editing the prompt.
- We can assume the user finished editing the prompt when the input loses focus.

Also note a similar issue results when typing in any other text field, like:
- Custom Options
- API Key
- Recording > Advanced Options > Global Options
- etc..

Leftium:
Actually, just binding all the inputs to the `onchange` event would be a better solution.

It seems the notifications are being spammed because the inputs are ~~bound to an event like `oninput` or `keydown`~~

Ah.. the inputs have Svelte reactive bindings. We can still switch these to just call the save action on the `onchange` event.

`onchange` vs `oninput` explained: https://ahmetustun.medium.com/understanding-the-difference-between-oninput-and-onchange-c7f2c1e0e728

Issue #1203, opened 2026-01-03
[plugin:@tailwindcss/vite:generate:serve] Invalid declaration: `true`


aspiers:
### Bug:
When I run a build of whispering from the latest git, I see an error:

```
[plugin:@tailwindcss/vite:generate:serve] Invalid declaration: `true`
```

<img width="1080" height="834" alt="Image" src="https://github.com/user-attachments/assets/6b73df6e-f8bb-4599-8815-b65ac25d19c0" />

### Expected behavior:

No error

### Steps to reproduce:
1. Build from latest `main` (currently 295e710b9555ec50ea1d898a0624854df126774c but I've seen this before)
2. Start up Whispering
3. Observe the error

### Additional context:
Running on openSUSE Tumbleweed Linux

### Platform

Linux

### Affected Component

UI-Web (web interface), UI-App (desktop interface)

### Whispering Version

v7.11.0

### Severity

Moderate (workaround available)

### Checklist

- [x] I have searched existing issues to ensure this hasn't been reported
- [x] I have included all relevant information above

Leftium:
#### I was unable to reproduce. Steps tried (from project root):
- Check out 295e710b9555ec50ea1d898a0624854df126774c
- `bun run nuke`
- `bun install`
- `cd apps/whispering && bun run dev:web`
 
Also repeated steps for current `main` branch (990d127bc077c58486809248dd190a90eb88ac8d)

Sometimes when there are large changes to the file system (from changing branches; pulling new changes) temporary build files are mixed with newly built files, which don't always work together. Often this causes errors like this.

There are two scripts that clean up cached built files. Both are run from Epicenter project root:
- `bun run clean`
- `bun run nuke`

The `clean` script should be sufficient in most cases. (`nuke` is more for Tauri build artifacts.)

Closing this issue; please let us know if you continue to see this error after running `clean` and even `nuke`.

More info: https://github.com/EpicenterHQ/epicenter/pull/1036

aspiers:
I did that and I got:

```
❯ cd apps/whispering && bun run dev:web
$ vite dev
/usr/bin/bash: line 1: vite: command not found
error: script "dev:web" exited with code 127
```

I assume `bun install` is needed first, so I did that and then got:

```
❯ bun run dev:web
$ vite dev
23:09:50 [vite] (client) Forced re-optimization of dependencies

  VITE v7.2.7  ready in 2127 ms
  ➜  Local:   http://localhost:1420/
[...]

PR #1248, opened 2026-01-13
fix(ui): suppress state_referenced_locally warnings with untrack()

Leftium:
## Summary

Silence Svelte 5 `state_referenced_locally` compiler warnings that were causing console noise and doubt from external devs. Confirmed these variables do not need to be reactive - behavior is correct as-is.

## Type of Change

- [x] `fix`: Bug fix

## Related Issue

Related to #1203

## Changes Made

Use Svelte's `untrack()` to explicitly mark intentionally non-reactive prop reads:

- `file-drop-zone.svelte`: wrap dev warning check (one-time mount validation)
- `toggle-group.svelte`: wrap setContext call (must run synchronously during init)
- `copy-button.svelte`: wrap size adjustment and clipboard instantiation
- `TranscriptionSelector.svelte`: wrap initial expanded state

### Desktop App Testing
- [ ] Tested on macOS
- [ ] Tested on Windows
- [ ] Tested on Linux
- [x] Not applicable (web-only change)

### General Testing
- [ ] Tested with multiple API providers (if applicable)
- [ ] Verified no API keys are exposed in logs or storage
- [x] Checked for console errors
- [ ] Tested on different screen sizes (if UI change)

## Checklist

- [x] My code follows the project's coding standards (see [CONTRIBUTING.md](../CONTRIBUTING.md))
- [x] I've used `type` instead of `interface` in TypeScript
- [x] I've used absolute imports where applicable
- [x] I've tested my changes thoroughly
- [ ] I've added/updated tests for my changes (if applicable)
- [x] My changes don't break existing functionality
- [ ] I've updated documentation (if needed)

## Screenshots/Recordings

N/A - no UI changes

## Additional Notes

The `state_referenced_locally` warning fires when reactive props are read outside reactive contexts. In these cases, capturing the initial value is intentional (setContext, one-time validation, static initialization).

### DIFF
--- a/apps/whispering/src/lib/components/settings/selectors/TranscriptionSelector.svelte
+++ b/apps/whispering/src/lib/components/settings/selectors/TranscriptionSelector.svelte
@@ -20,6 +20,7 @@
 	import SettingsIcon from '@lucide/svelte/icons/settings';
 	import ChevronRightIcon from '@lucide/svelte/icons/chevron-right';
 	import { SvelteSet } from 'svelte/reactivity';
+	import { untrack } from 'svelte';
 
 	let { class: className }: { class?: string } = $props();
 
@@ -41,7 +42,9 @@
 	);
 
 	const selfHostedServices = $derived(
-		TRANSCRIPTION_SERVICES.filter((service) => service.location === 'self-hosted'),
+		TRANSCRIPTION_SERVICES.filter(
+			(service) => service.location === 'self-hosted',
+		),
 	);
 
 	const localServices = $derived(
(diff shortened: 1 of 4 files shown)


braden-w:
Thanks for tackling these warnings! However, after researching this, `untrack()` at top-level script doesn't actually do anything; it only has an effect inside `$effect` or `$derived` blocks.

The correct approaches are:
1. For intentional one-time initialization: use `// svelte-ignore state_referenced_locally` 
2. For setContext: use getter functions so context stays reactive: `{ get variant() { return variant } }`
3. For values that should track changes: use `$derived`

Would you be open to reworking this with those patterns instead?

braden-w:
I've created #1257 with an alternative approach. After researching this more, `untrack()` at top-level script doesn't actually have any effect; it only works inside `$effect` or `$derived` blocks.

The proper fixes are:
- `svelte-ignore state_referenced_locally` for intentional one-time initialization (TranscriptionSelector, copy-button, file-drop-zone)
- Getters for context values that need to stay reactive (toggle-group)

I've added you as co-author since your PR identified these warnings. Going to merge that one instead, but thank you for surfacing this!

PR #1236, opened 2026-01-10
fix(whispering): save text inputs on blur instead of every keystroke

Leftium:
### Problem
Editing text inputs in settings (System Prompt, API keys, etc.) triggered a save on **every keystroke**, resulting in constant toast notifications.

### Solution
Changed text inputs to save **on blur** instead of on every keystroke:

- Use local `$state()` for input values
- Trigger `settings.updateKey()` on `onchange` (fires on blur/enter)
- For FFmpeg settings: added `commitChanges` callback prop so child component can signal when to save

### Result
Users can type freely without toast spam. A single toast appears when they leave the field.

---

## Type of Change

- [x] `fix`: Bug fix

---

## Related Issue

Closes #1214

---

### System Prompt (main issue)
1. Navigate to **Settings &#x2192; Transcription**
2. Edit the **System Prompt** field
3. Confirm that:
   - No toasts while typing
   - Single toast appears when you click outside or press Tab

### API Keys
- Test any API key input (Settings &#x2192; API Keys or select a cloud transcription service)
- Same behavior: no toast while typing, single toast on blur

### FFmpeg Options

**Setup:**
1. Go to Settings &#x2192; Recording
2. Set Recording Mode to "Manual"
3. Set Recording Method to "FFmpeg"

**Test Cases:**

1. **Select dropdowns (Format, Sample Rate, Bitrate/Quality)**
   - Change format from WAV &#x2192; MP3
   - Verify: Toast appears once, preview updates, setting persists on page refresh

2. **Advanced text inputs (Global/Input/Output Options)**
   - Expand "Advanced Options"
   - Type multiple characters in Global Options field
   - Verify: No toast while typing
   - Click outside (blur) or press Tab
   - Verify: Single toast appears, setting persists on refresh

3. **Reset buttons**
   - Modify Global Options text
   - Click the reset button (&#x21ba;)
   - Verify: Single toast appears, value resets to default

4. **"Reset all" button**
   - Modify multiple fields
   - Click "Reset all" link
   - Verify: Single toast appears, all values reset

5. **Live preview**
   - Type in any text input
   - Verify: Command preview updates in real-time while typing (before blur)

---

## Comparison to #1234

| Approach | This PR (blur) | #1234 (debounce) |
|----------|----------------|------------------|
| When saves | On blur/enter | 500ms after typing stops |
| New dependencies | None | Debounce utility |
| Files changed | 12 | 4 |
| Scope | All text inputs | System prompt only |

PR #1255, opened 2026-01-20
fix(settings): remove success toast for routine updates

braden-w:
Every settings change was triggering a toast AND OS notification, causing spam when typing in settings fields. This removes the `onUpdateSuccess` callback that fired on every settings update.

The change is minimal: just removing the success notification callback from `createPersistedState` in `settings.svelte.ts:54-59`. Error notifications remain intact for actual failures, and the explicit success toast for `switchRecordingMode` is preserved since mode switching is a significant user action (not routine typing).

Related to #1236


braden-w:
Closing to review changes before PR creation

PR #1257, opened 2026-01-20
fix(ui): properly fix state_referenced_locally warnings

braden-w:
Fixes Svelte 5 `state_referenced_locally` warnings using the correct patterns instead of `untrack()` (which has no effect at top-level script).

**Fixes applied:**

| File | Fix | Reasoning |
|------|-----|-----------|
| TranscriptionSelector.svelte | `svelte-ignore` | One-time init to expand selected service; re-tracking would reset user's expand/collapse |
| copy-button.svelte | `svelte-ignore` (2x) | Size adjustment and clipboard instance are intentionally static |
| file-drop-zone.svelte | `svelte-ignore` | Dev warning should only fire once at init |
| toggle-group.svelte | **Getters** | Context must stay reactive for child components |

The key insight: `untrack()` only works inside `$effect` or `$derived` blocks. At top-level, it does nothing. For intentional one-time initialization, `svelte-ignore` is the honest approach. For context that needs to stay reactive, use getters.

Related to #1248

Co-authored-by: John-Kim Murphy <Leftium@users.noreply.github.com>

### DIFF
--- a/apps/whispering/src/lib/components/settings/selectors/TranscriptionSelector.svelte
+++ b/apps/whispering/src/lib/components/settings/selectors/TranscriptionSelector.svelte
@@ -41,7 +41,9 @@
 	);
 
 	const selfHostedServices = $derived(
-		TRANSCRIPTION_SERVICES.filter((service) => service.location === 'self-hosted'),
+		TRANSCRIPTION_SERVICES.filter(
+			(service) => service.location === 'self-hosted',
+		),
 	);
 
 	const localServices = $derived(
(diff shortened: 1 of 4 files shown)


Leftium:
@braden-w This PR looks good. I think the other PR was also valid, too:
- `untrack()` took an already untracked variable (thus the warning) and made it untracked. While this looks like a NOP, it signals to both the compiler and human that this untracking is intentional. This syntax is more compact, and sometimes `svelte-ignore` does not work or results in other lint errors.
- `setToggleGroupCtx` seemed to be reactive via other means, so it did not need to be made reactive here. The ToggleGroup behavior seems to be the same in both PRs:
  - From the Transcription model quick settings, open up a group without selecting an item from the group:
    -  This group remains open when the transcription model quick settings are opened again
    - *Except* when navigating to and from the home page: the group with the currently selected model is always toggled open.
    - Not sure what the expected behavior is, now...

PR #1262, opened 2026-01-21
fix: improve clean script to clear Tauri webview and Vite caches

Leftium:
fix: improve clean script to clear Tauri webview and Vite caches

## Summary
- Cleans additional folders to fix stale cache issues that cause `lifecycle_outside_component` errors and blank pages after switching branches.
- Added instructions on how to really clear cache in Chrome (not obvious)
- Made cross-platform

## Problem
The previous `clean` command missed several caches that caused hard-to-debug issues:
- Tauri webview cache (caused `lifecycle_outside_component` Svelte errors)
- Vite dependency cache (`504 Outdated Optimize Dep` errors)
- Browser cache (requires manual clearing - now documented)

## Changes
- Add Tauri webview cache clearing (cross-platform: macOS, Linux, Windows)
- Add Vite cache clearing (`node_modules/.vite`)
- Add browser cache clearing instructions in output
- Make script cross-platform (replaces `rm -rf` with Node fs API)
- Dynamically discover all workspace directories (no more hardcoded paths)
- Prompt to run `bun install` after clean (default: Y)
- Add `--nuke` flag for removing Rust target directory

## Usage
```bash
bun run clean  # Clean JS artifacts + caches, prompt to reinstall
bun run nuke   # Above + remove Rust target (expensive!)
```

### DIFF
--- a/.claude/skills/github-issues/SKILL.md
+++ b/.claude/skills/github-issues/SKILL.md
@@ -3,7 +3,14 @@ name: github-issues
 description: GitHub issue comment guidelines for community interaction. Use when responding to GitHub issues, bug reports, feature requests, or any GitHub discussion.
 ---
 
-# GitHub Issue Comment Guidelines
+# GitHub Issue & PR Comment Guidelines
+
+## Anti-Patterns (Avoid These)
+
+- **Over-structured responses**: Don't use headers, numbered sections, or bullet lists for simple replies. A conversational paragraph is usually better.
+- **Formulaic openings**: Don't start every comment identically. Match the tone to the conversation.
+- **Restating what's obvious**: If someone asked a question, just answer it. Don't recap what they said.
+- **Corporate announcements**: "We are pleased to announce..." — just say what changed.
 
 ## Opening Pattern
 
@@ -96,7 +103,21 @@ Thank you!
 Hey @username, sorry to hear [problem]! Did you ever get a fix?
 ```
 
+### PR Discussion (back-and-forth)
+
+```
+Good catch! Updated to only clear the dev app cache in nuke mode.
+
+The flow is now:
+- `bun clean`: artifacts and node_modules
+- `bun nuke`: above + Rust targets + dev cache
+
+Let me know if you want a confirmation prompt before clearing.
+```
+
 ## Writing Style Notes
+
+See [writing-voice](../writing-voice/SKILL.md) for punctuation and tone guidelines.
 
 - Use casual, approachable language
 - Be genuinely enthusiastic about user contributions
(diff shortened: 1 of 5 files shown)


braden-w:
Heads up: I removed the Tauri webview cache clearing (commit 60ff8e9) because the WebKit directories contain localStorage data — that's where all the API keys and settings live. Nuking that would wipe user data.

Should we still clear cache for `com.tauri.dev` (the dev app) while leaving `com.bradenwong.whispering` alone? Dev app probably doesn't have anything important.

Leftium:
One of the main purposes of the `clean`/`nuke` script is to get the Whispering app back to a working condition. So it is very probable users of the script cannot do anything in-app. (I guess I *do* personally run the script just to make sure I'm not testing cached code.)

How about moving Tauri webview cache clearing to the `nuke` option? (Or maybe even more extreme level than `nuke`)

We could gate Tauri webview clearing behind both the `nuke` flag and a prompt that defaults to "No." I would not expect this most extreme level to be used often; just when nothing else seems to work.

BTW here is the issue where another dev has an issue that I suspect is related to caching issues like this: https://github.com/EpicenterHQ/epicenter/issues/1203

Leftium:
Some more context for the new items in the clean script PR:

- After switching from the overlay PR branch back to main, Whispering would not work:
  - Blank screen on load; both desktop and web versions.
  - Errors in browser dev console.
- I tried several things that did not help:
  - restart dev server
  - bun clean
  - bun nuke
  - switch back to PR branch and back to main
  - tried checking out commits I knew had been working just hours before
- Notably, Chrome browser "incognito" mode worked. But just for the web version in "incognito" mode.
- I added these items because clearing them restored Whispering to a working condition in this case:
  - clearing Tauri cache (just desktop version)
  - "Empty cache and hard reload" from right-click Chrome reload button (just web version):
  - (I don't think Vite cache helped, but it seemed worth adding, too)
image

Leftium:
> Should we still clear cache for [com.tauri.dev](http://com.tauri.dev/) (epicenter - dev app) while leaving com.bradenwong.whispering (production app) untouched?

Ah, I didn't realize the production app data was getting cleared too. We should just clear the dev data.

braden-w:
(Apologies for the previous comment; accidentally posted a coding agent response 😅)

Good catch on the prod vs dev distinction! Updated the script to only clear the dev app cache (`com.tauri.dev`) in nuke mode, leaving production data untouched.

The flow is now:
- `bun clean` — build artifacts, caches, node_modules
- `bun nuke` — above + Rust targets + dev webview cache

This should handle the caching issues you hit when switching branches without nuking anyone's API keys.

Leftium:
Unfortunately, I am still getting errors after switching branches. This time:
- switched from branch for PR 1293 to branch for PR 1292
- ran `bun run clean`; when errors persisted `bun run nuke` (it seems the updated scripts ran, based on the terminal output)
- (dev server was shut down/restarted because clean/nuke scripts were run from same terminal)
- **update:** the local web server works (http://localhost:1420). So this time problem is only with Tauri.
image

PR #1272, opened 2026-01-23
feat(whispering): add ElevenLabs scribe_v2 model support

tatn:
This adds support for ElevenLabs' latest scribe_v2 transcription model. The v2 model offers improved accuracy (97% vs 96.7% for v1), speaker diarization for up to 48 speakers, entity detection, keyterm prompting, and dynamic audio tagging. According to ElevenLabs documentation, scribe_v1 is "outclassed by v2 models."

The new model is added to the ELEVENLABS_TRANSCRIPTION_MODELS array as the first entry, making it the recommended option in the UI. The scribe_v1 description has been updated to indicate it's the "previous generation" model. Existing users will continue using their selected model since the default in settings.ts remains unchanged.

Documentation in the README has been updated to list scribe_v2 alongside the existing models.


braden-w:
LGTM! Thank you :)

Xiller:
Hi, i installed Windows Version 7.11.0 and there is only v1 and v1 experimental. Everything else: Awesome!

PR #1293, opened 2026-01-31
fix: improve global shortcut recording UX

sudhanshu112233shukla:
### Problem
Users couldn't enter complex shortcuts (like Ctrl+Option+Cmd+R) on macOS because:
1. Local shortcuts (e.g. 'R') triggered while recording
2. Capture window (300ms) was too short for 3 modifiers

### Solution
1. **Recording Mode:** Added flag to disable local shortcuts while recording (`isRecordingShortcut`)
2. **Better Capture:** Increased window to 600ms and added live preview of pressed keys

### Testing
- Validated via static analysis (TypeScript check passed)
- Manual testing required on macOS

### DIFF
--- a/apps/whispering/src/lib/services/isomorphic/local-shortcut-manager.ts
+++ b/apps/whispering/src/lib/services/isomorphic/local-shortcut-manager.ts
@@ -34,6 +34,21 @@ const shortcuts = new Map<
 >();
 
 /**
+ * Global flag to track if we're currently recording a keyboard shortcut.
+ * When true, all local shortcuts are disabled to prevent interference.
+ */
+let isRecordingShortcut = false;
+
+/**
+ * Set the recording mode state. When recording is enabled, local shortcuts
+ * will not fire, allowing users to record any key combination without
+ * triggering existing shortcuts.
+ */
+export function setRecordingMode(recording: boolean) {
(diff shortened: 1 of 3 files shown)


sudhanshu112233shukla:
@braden-w do check it it shows any more error on mac i will fix that 
thank you!!

Leftium:
I think this PR makes it easier to enter a shortcut with `Ctrl+Option+Cmd`. But it is still difficult:
- 600ms still seems too short
- When I realized 600ms is still too short, I tried entering *very* quickly. This helped.
  - When I naturally enter the shortcut, I hold `Ctrl+Option+Cmd`, (slight pause), then press `R`
  - The UI suggests pressing in reverse order on MacOS: Hold  `R`, (slight pause), then `Ctrl+Option+Cmd` 
- `Ctrl+Option+Cmd+R` gets another type of error (see screen shot). (`Ctrl+Option+Cmd+T` does not seem to get this error as often.)

image

sudhanshu112233shukla:
Yup you are right, give me sometime let me see what I can do to fix this !!!

I will text you within hr

sudhanshu112233shukla:
@Leftium  i am working on windows  so I want you to help me check it on mac please

PR #1425, opened 2026-02-25
feat(linux): add ydotool support for paste on Wayland

srix:
## Summary

- Adds `system.useYdotool` setting for Linux Wayland users where `enigo` cannot simulate keystrokes into other windows
- When enabled, uses `ydotool key ctrl+v` to simulate paste instead of enigo, enabling paste in native Wayland apps (Konsole, VS Code, Cursor, KWrite, etc.)
- Falls back to enigo when the setting is disabled or ydotool is unavailable
- Also updates `tauri-plugin-http` to 2.5.7 and adds app data paths to the asset protocol scope for WAV file playback on Linux

## Configuration

The `system.useYdotool` setting defaults to `false`. Linux Wayland users who need this should:

1. Install ydotool: `sudo apt install ydotool`
2. Set `system.useYdotool` to `true` in their settings (`whispering-settings` in localStorage)

No UI toggle has been added yet. When the setting is `false` (default), the original `enigo` behavior is unchanged on all platforms.

## Context

On Linux with Wayland (especially KDE Plasma), `enigo` cannot inject keyboard events into other windows due to Wayland's security model. `wtype` also doesn't work on KDE as it lacks `wlr-virtual-keyboard` protocol support. `ydotool` works at the kernel level via `/dev/uinput`, making it compatible with all compositors and apps.

## Test plan

- [x] Verify paste works on Linux Wayland (KDE) with setting enabled and ydotool installed
- [x] Verify paste still works on macOS and Windows (no behavior change, setting defaults to false)
- [x] Verify paste works on Linux X11 with setting disabled (enigo, default behavior)
- [x] Verify `simulate_enter_keystroke` works with ydotool enabled

### DIFF
--- a/apps/whispering/src-tauri/src/lib.rs
+++ b/apps/whispering/src-tauri/src/lib.rs
@@ -209,7 +209,7 @@ use tauri_plugin_clipboard_manager::ClipboardExt;
 /// This approach is faster than typing character-by-character and preserves
 /// the user's clipboard, making it ideal for inserting transcribed text.
 #[tauri::command]
-async fn write_text(app: tauri::AppHandle, text: String) -> Result<(), String> {
+async fn write_text(app: tauri::AppHandle, text: String, use_ydotool: Option<bool>) -> Result<(), String> {
     // 1. Save current clipboard content
     let original_clipboard = app.clipboard().read_text().ok();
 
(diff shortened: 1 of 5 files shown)


srix:
## Updated for latest main (April 2026)

This PR was originally created in February against an older codebase. I've rebased and re-applied the changes to match the current architecture:

### What changed in this update

**Rust (`lib.rs`)** — unchanged from the original PR:
- `write_text` and `simulate_enter_keystroke` accept an optional `use_ydotool` parameter
- On Linux, if enabled, uses `ydotool key ctrl+v` / `ydotool key Return` instead of enigo
- Falls back to enigo if ydotool fails or isn't installed

**TypeScript — adapted to new codebase structure:**
- **`workspace/definition.ts`**: Added `system.useYdotool` as a workspace KV entry (`defineKv(type('boolean'), false)`) since the old `settings.ts` Arktype schema was removed
- **`services/text/desktop.ts`**: Passes `useYdotool` to the Rust `invoke` calls for `write_text` and `simulate_enter_keystroke`
- **`services/text/index.ts`**: Wires the KV setting into the desktop text service via `getUseYdotool` dependency

**New — Settings UI toggle:**
- Added a "Use ydotool (Linux Wayland)" switch in Settings > General
- Only visible on Linux desktop (guarded by `PLATFORM_TYPE === 'linux'` and `window.__TAURI_INTERNALS__`)

### Why ydotool?
On Linux Wayland compositors, enigo cannot simulate keypresses (Ctrl+V, Enter) because Wayland isolates input between applications. ydotool talks directly to `/dev/uinput` and works regardless of the display server.

uberkael:
What does prevent this from being merged?

also:
enigo (v0.6.1) added restore_token support here https://github.com/enigo-rs/enigo/pull/532
maybe still not working https://github.com/enigo-rs/enigo/pull/536

PR #1374, opened 2026-02-18
fix(sound): release Audio elements after playback to prevent PipeWire node leaks

aspiers:
Each `HTMLAudioElement` registers a PipeWire output node that persists as long as the element is reachable. Previously, `Audio` objects were created eagerly at module load time and cached indefinitely, so every sound that played left a permanent PipeWire node behind — accumulating as multiple Whispering entries in PipeWire audio managers:

```
┌──────────────────────────────────────────────────────────┐
│  Before: eager construction + indefinite caching         │
│                                                          │
│  Module load  ──→  new Audio() × N  ──→  PipeWire nodes │
│                    (cached forever)      (never released) │
│                                                          │
│  HMR reload   ──→  new Audio() × N  ──→  more nodes     │
│                    (old ones orphaned)                    │
└──────────────────────────────────────────────────────────┘

┌──────────────────────────────────────────────────────────┐
│  After: on-demand creation + release on ended            │
│                                                          │
│  playSound()  ──→  new Audio()  ──→  play  ──→  ended   │
│                                        │         │       │
│                                        │    el.src = ''  │
│                                        │    el.load()    │
│                                        │         │       │
│                                        ▼         ▼       │
│                                   PipeWire node removed  │
└──────────────────────────────────────────────────────────┘
```

The `playSound()` function creates a fresh `Audio` element per call, waits for the `ended` event, then clears the `src` and calls `load()` to tear down the underlying media resource and its PipeWire node.

Fixes #842

PR #1548, opened 2026-03-21
fix(sound): use Web Audio API to fix PipeWire node accumulation and volume resets

aspiers:
## Summary

Replaces `HTMLAudioElement` with the Web Audio API (`AudioContext` + `BufferSourceNode`) for sound playback. This fixes both the PipeWire node accumulation from #842 and a volume reset issue on Linux where WirePlumber restores full volume on every sound play.

Supersedes #1374.

## Problem

`HTMLAudioElement` has two issues on Linux with PipeWire/WirePlumber:

1. **Node accumulation (#842):** Each `Audio` element registers a PipeWire output node that persists as long as the element is reachable. The original code cached 8 elements at module load, and HMR reloads orphaned old elements without releasing their nodes.

2. **Volume resets:** #1374 fixed accumulation by creating a fresh element per playback and tearing it down after `ended`. However, WebKit tears down and recreates the underlying PipeWire media pipeline on every new `Audio()` or `src` change, so WirePlumber sees each playback as a brand-new stream and restores it to its database volume (1.0 / full) before the user's adjustment can be saved back.

```
HTMLAudioElement approaches — both broken:

Cached elements:   Module load → new Audio() × 8 → PipeWire nodes pile up on HMR
Per-play elements: playSound() → new Audio() → WirePlumber restores 1.0 → volume resets
```

## Solution

A single `AudioContext` is created lazily on first use and kept alive for the app's lifetime. This registers exactly one PipeWire node that persists across all sound playback. Audio files are decoded once into `AudioBuffer`s (cached in memory), and each `playSound()` call creates a lightweight `BufferSourceNode` — no new PipeWire nodes.

```
AudioContext approach:

First playSound() → new AudioContext() → 1 PipeWire node (persists forever)
                  → fetch + decodeAudioData → cached AudioBuffer

Subsequent calls  → new BufferSourceNode (no PipeWire overhead)
                  → connect to destination → start()
                  → auto-cleaned up after playback
```

- No node accumulation — one stable node for the app's lifetime
- WirePlumber can reliably track and persist volume adjustments
- Decoded buffers are cached, so repeat plays are efficient
- No `ended` event race conditions that could hang the playback promise

Fixes #842

### DIFF
--- a/apps/whispering/src/lib/services/sound/assets/index.ts
+++ b/apps/whispering/src/lib/services/sound/assets/index.ts
@@ -10,13 +10,63 @@ import cancelSoundSrc from './zapsplat_multimedia_click_button_short_sharp_73510
 import transformationCompleteSoundSrc from './zapsplat_multimedia_notification_alert_ping_bright_chime_001_93276.mp3';
 import transcriptionCompleteSoundSrc from './zapsplat_multimedia_ui_notification_classic_bell_synth_success_107505.mp3';
 
-export const audioElements = {
-	'manual-start': new Audio(startManualSoundSrc),
-	'manual-cancel': new Audio(cancelSoundSrc),
-	'manual-stop': new Audio(stopManualSoundSrc),
-	'vad-start': new Audio(startVadSoundSrc),
-	'vad-capture': new Audio(captureVadSoundSrc),
-	'vad-stop': new Audio(stopVadSoundSrc),
-	transcriptionComplete: new Audio(transcriptionCompleteSoundSrc),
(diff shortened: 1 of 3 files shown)


aspiers:
CI failure is pre-existing on `main` — not caused by this PR. The lint errors are in unrelated files (`apps/api/`, `apps/fuji/`, `apps/honeycrisp/`, `packages/workspace/`) and include a biome schema version mismatch (2.4.6 vs 2.4.8). All tests pass locally (557 pass, 0 fail).

PR #1620, opened 2026-04-07
refactor(cli): remove jsrepo dependency and install command

braden-w:
The CLI's only use of jsrepo was `epicenter install`, which fetched workspace source files from a jsrepo registry into `./workspaces/` and wired them into `epicenter.config.ts`. In practice, no registry exists—every app in the monorepo defines workspaces inline via `@epicenter/workspace` imports. The install command was speculative infrastructure for a distribution model that hasn't materialized.

For now, workspace definitions will be distributed as normal npm packages. Factory functions like `createTabManagerWorkspace` can be exported from existing packages and composed with `.withActions()` at the call site. This is simpler and gives us versioning, lockfiles, and `bun update` for free.

```bash
# Before — jsrepo registry fetch
epicenter install github/epicenterhq/workspaces/my-notes
#   → resolveRegistries() → parseWantedItems() → fetch → write files
#   → patch epicenter.config.ts

# After — normal npm package
bun add @epicenter/my-notes-workspace
```

```typescript
// epicenter.config.ts — just import the package and compose
import { createMyNotes } from '@epicenter/my-notes-workspace';
import { filesystemPersistence } from '@epicenter/workspace/extensions/persistence/sqlite';

export const myNotes = createMyNotes()
  .withExtension('persistence', filesystemPersistence({ filePath: './data/my-notes.db' }));
```

### What was removed

```typescript
// packages/cli/src/commands/project.ts
// — the entire jsrepo resolve → fetch → write pipeline:
import { resolveRegistries, parseWantedItems, resolveWantedItems, resolveAndFetchAllItems } from 'jsrepo';

async function installWorkspace(dir: string, itemArg: string, registryFlag?: string) {
  const registriesResult = await resolveRegistries([registryUrl], { cwd, providers: DEFAULT_PROVIDERS });
  const parsedResult = parseWantedItems([itemName], { ... });
  const resolvedResult = await resolveWantedItems(parsedResult.value.wantedItems, { ... });
  const fetchedItems = await resolveAndFetchAllItems(resolvedResult.value);
  // write files to ./workspaces/<name>/, patch epicenter.config.ts
}

// packages/cli/src/util/update-config.ts
// — addWorkspaceToConfig(), toFactoryName() (only used by install)

// packages/cli/package.json
// — "jsrepo": "catalog:" dependency
```

### What remains

- `epicenter init` — scaffolds `epicenter.config.ts`, `package.json`, `.gitignore`
- `epicenter uninstall` — removes `./workspaces/<id>/` and patches config (cleaned up in #1621)
- `packages/ui/jsrepo.config.ts` — unrelated; manages shadcn-svelte-extras component installation

### Reinstating jsrepo later

If the ecosystem grows to where third-party workspace authors need the shadcn-style "own the code" pattern (copy source into project, interactive diff updates), jsrepo can be reintroduced. The commit being removed here documents the full integration pattern: `resolveRegistries` → `parseWantedItems` → `resolveWantedItems` → `resolveAndFetchAllItems` → write files → patch config.

Follow-up: #1621 removes the now-dead `uninstall` command and `update-config.ts` that only existed to support this flow, and fixes the `CONFIG_TEMPLATE` to show the npm pattern.

6 files changed, -278 lines.

### DIFF
--- a/apps/dashboard/src/lib/api.ts
+++ b/apps/dashboard/src/lib/api.ts
@@ -30,19 +30,12 @@ import { defineErrors, extractErrorMessage } from 'wellcrafted/error';
 import { type Result, tryAsync } from 'wellcrafted/result';
 import { auth } from './auth';
 
-// Re-export only the contract types that components actually import.
-// Components needing other types can import from @epicenter/api/billing-contract directly.
-export type {
-	AttachResponse,
-	PortalResponse,
-	PreviewResponse,
-} from '@epicenter/api/billing-contract';
 
 /**
- * Tagged errors for the billing API boundary.
+ * Tagged error for the billing API boundary.
  *
- * Two variants cover the two failure modes: the HTTP request itself
- * failed (network, timeout, DNS) or the server returned a non-OK status.
+ * Covers both network failures (fetch throws) and non-OK HTTP
+ * responses (our status guard throws).
  */
 export const BillingApiError = defineErrors({
 	RequestFailed: ({ path, cause }: { path: string; cause: unknown }) => ({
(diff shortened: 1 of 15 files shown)


braden-w:
Follow-up: #1621 removes the now-dead `uninstall` command and `update-config.ts` module that only existed to support the jsrepo install flow, and fixes the `CONFIG_TEMPLATE` to show the npm package pattern instead of the defunct `./workspaces/` path. Merge this first, then #1621.

PR #1627, opened 2026-04-07
refactor(ai): collapse actionsToClientTools + toToolDefinitions into actionsToAiTools

braden-w:
Both callsites in the repo—tab-manager and opensidian—always called `actionsToClientTools` then immediately fed the result into `toToolDefinitions`. The intermediate array was also used directly by `ChatClient`, so both outputs were needed, but the two-step pipeline was choreography every consumer had to remember:

```typescript
// Before: two-step pipeline
const tools = actionsToClientTools(workspace.actions);
const defs = toToolDefinitions(tools);
```

```typescript
// After: single call, destructure both
const { clientTools, definitions } = actionsToAiTools(workspace.actions);
```

Collapsing into one function also eliminates a type-widening round-trip. The old `toToolDefinitions` accepted `readonly AnyClientTool[]` (the TanStack AI base type), which widened `inputSchema` to `SchemaInput`—forcing `as JSONSchema` casts to narrow it back. The new function derives definitions directly from the action entries, so the cast goes straight from `action.input` (TypeBox `TSchema`) to `JSONSchema` without the intermediate widening.

The name change from `actionsToClientTools` to `actionsToAiTools` drops TanStack-specific jargon ("client tools") from the public API surface. Consumers importing from `@epicenter/ai` don't need to know about TanStack's internal type hierarchy—they need "give me AI tools from my actions."

All 4 existing tests pass under the new API, restructured into `clientTools` and `definitions` describe blocks.

### DIFF
--- a/apps/opensidian/src/lib/client.ts
+++ b/apps/opensidian/src/lib/client.ts
@@ -1,4 +1,4 @@
-import { actionsToClientTools, toToolDefinitions } from '@epicenter/ai';
+import { actionsToAiTools } from '@epicenter/ai';
 import { APP_URLS } from '@epicenter/constants/vite';
 import { createSqliteIndex, createYjsFileSystem } from '@epicenter/filesystem';
 import { createSkillsWorkspace } from '@epicenter/skills';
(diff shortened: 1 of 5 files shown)


braden-w:
Superseded — folding rename + restructure into a single clean-break PR.

PR #1672, opened 2026-04-16
fix(whispering): handle quick-release push-to-talk correctly

samiralibabic:
## Summary
- Queue manual stop when push-to-talk release happens during startup
- Consume queued stop immediately after startup completes
- No-op manual start if recorder is already recording

Fixes #1671.

## Testing
- previously verified quick tap no longer leaves recording latched on
- previously verified hold-to-talk still works
- previously verified second tap no longer triggers another broken start-only path
- previously verified toggle recording behavior remains unchanged
- after rebasing onto latest `main`, `bun run typecheck` in `apps/whispering` still fails on current upstream-level TypeScript/Svelte issues unrelated to this PR

## Changelog
- Fix push-to-talk getting stuck when the shortcut is tapped quickly on macOS

PR #1674, opened 2026-04-16
fix(whispering): restore transformation execution flow

samiralibabic:
## Problem
Transformations were failing in multiple user-visible ways:

- the transformation test pane could hit the provider but return no output
- clipboard-triggered transform flows could fail before surfacing a useful error
- unexpected provider/runtime exceptions during step execution could bypass the normal fail path
- desktop HTTP calls could fail with `fetch_read_body missing required key streamChannel` when the Rust and JS Tauri HTTP plugin versions drifted

This branch is stacked on #1672 in Git history, so GitHub may show the quick-release commits in this PR until #1672 lands.

## Root Causes
- Clipboard-triggered transform flows called `transformInput` without passing `steps`.
- Runtime/provider exceptions during step execution were not normalized into failed transformation runs.
- OpenAI-compatible providers could return response shapes that were not handled consistently.
- The Rust `tauri-plugin-http` version drifted from the resolved JS plugin version.

## What This PR Changes
- Pass transformation `steps` through clipboard-triggered transformation flows.
- Catch provider/runtime failures during step execution and persist them as failed runs instead of leaving runs stuck in `running`.
- Normalize OpenAI-compatible response handling for transformation output extraction.
- Pin the Rust HTTP plugin to `2.5.8` to match the resolved JS plugin version and restore desktop transformation API calls.

## Verification
- `cargo check` in `apps/whispering/src-tauri` passes.
- `bun run typecheck` in `apps/whispering` still fails on current upstream-level TypeScript/Svelte issues unrelated to this PR.

## Notes
- During rebase onto latest `main`, the older DB-service transformation run cache changes from the original PR became obsolete because the current codebase now stores transformation runs in workspace tables. Those stale `services.db` changes were intentionally not reintroduced.


samiralibabic:
Closing this as superseded by #1903, which keeps only the confirmed current transformation-run failure normalization fix.

PR #1695, opened 2026-04-17
fix(workspace): constrain materializer TableRow with BaseRow intersection

braden-w:
## Why

The markdown materializer's local `TableRow<TName>` type alias was a deferred conditional type—TypeScript couldn't resolve it when `TName` was still generic, so `.id` access and object spread both failed with TS2339/TS2698.

## What changed

**1 file, 2 lines.**

```typescript
// Before — deferred conditional, .id and spread fail
type TableRow<TName extends keyof TTables & string> =
    TTables[TName] extends TableHelper<infer TRow> ? TRow : never;

// After — intersection gives TS the structural guarantee
type TableRow<TName extends keyof TTables & string> =
    (TTables[TName] extends TableHelper<infer TRow> ? TRow : never) & BaseRow;
```

The `& BaseRow` intersection is always truthful because every table row must extend `BaseRow` (which has `id: string` and `_v: number`). It just makes the compiler aware of what the runtime already guarantees.

Fixes all 4 type errors in `materializer.ts` (2× TS2339, 2× TS2698).

### DIFF
--- a/packages/workspace/src/extensions/materializer/markdown/materializer.ts
+++ b/packages/workspace/src/extensions/materializer/markdown/materializer.ts
@@ -1,7 +1,7 @@
 import { mkdir, readdir, readFile, unlink, writeFile } from 'node:fs/promises';
 import { join } from 'node:path';
 import type { MaybePromise } from '../../../workspace/lifecycle.js';
-import type { KvHelper, TableHelper } from '../../../workspace/types.js';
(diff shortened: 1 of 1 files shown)


braden-w:
Closing this as obsolete.

The old `packages/workspace/src/extensions/materializer/...` path no longer exists, and current table inference already guarantees `BaseRow` on rows through `InferTableRow`. The constraint this PR added has landed in the newer table-layer shape.

Reference: `packages/workspace/src/document/table.ts`.

PR #1699, opened 2026-04-20
Improve FFmpeg recording stability on Windows

YazzinDev:
# Improve FFmpeg recording stability on Windows

## Overview
This PR addresses critical stability issues for FFmpeg-based recording on Windows. It resolves "Malformed dshow input string" errors during initialization and "Recording file is empty" errors during the stopping phase.

## Changelog
- Fix initialization errors for microphones with colons in their names on Windows
- Fix "Recording file is empty" error by ensuring audio data is correctly flushed to disk when stopping

### Robust Device Identification (Frontend)
- Updated `parseDevices` to extract and use the DirectShow "Alternative name" (unique hardware path) as the stable device identifier on Windows.
- Implemented automatic colon escaping (`\:`) in device identifiers to prevent FFmpeg parsing errors in the `dshow` input.
- Refactored the Windows device parser to support multi-line output from FFmpeg.

### Graceful Process Termination (Backend & Frontend)
- Implemented `stop_command` in the Rust backend to send a `'q'` character to FFmpeg's `stdin`, triggering a graceful flush of audio buffers.
- Added a global process management map in Rust to track and interact with active child processes.
- Updated the frontend to explicitly wait for process termination before attempting to read the resulting audio file.

### Backend Fixes & Hygiene
- Resolved Windows-specific compilation errors and type mismatches in `command.rs`.
- Added test cases verifying multi-line parsing and colon escaping.

## Verification Results
- Verified that devices with colons in their names (e.g., "Virtual Audio Device (With:Colon)") initialize correctly.
- Confirmed that 5-second recordings result in files with actual audio data (~81KB) rather than 0 bytes.
- Confirmed that no orphaned FFmpeg processes remain in the background after stopping.

### DIFF
--- a/apps/whispering/src-tauri/src/command.rs
+++ b/apps/whispering/src-tauri/src/command.rs
@@ -1,5 +1,9 @@
 use serde::{Deserialize, Serialize};
-use std::process::{Command, Stdio};
+use std::collections::HashMap;
+use std::io::Write;
+use std::process::{Child, Command, Stdio};
+use std::sync::Mutex;
+use lazy_static::lazy_static;
 
 #[cfg(target_os = "windows")]
 use std::os::windows::process::CommandExt;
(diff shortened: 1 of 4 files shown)


braden-w:
Hey @YazzinDev, thanks for the careful Windows work here!

The direction changed underneath this PR. Whispering is going through a lot of audio-pipeline cleanup right now, and current main no longer has an FFmpeg recording backend. That means the DirectShow device parsing, stdin stop command, and FFmpeg child-process management this PR improves are no longer on the active recording path.

The cleanup chain is:

```txt
#1816  remove FFmpeg as a recording method
#1826  replace FFmpeg local decode with Symphonia + libopus
#1828  replace FFmpeg upload compression with Rust Opus
```

Because this PR fixes a removed path, I am going to close it instead of asking for a rebase. That is not a knock on the patch. I appreciate the care you put into the Windows edge cases here, especially the DirectShow device names and process shutdown behavior.

If a similar Windows recording issue shows up in the new path, I would be happy to look at a fresh patch there. Thank you!

PR #1702, opened 2026-04-24
feat(ui): adopt shadcn-svelte luma preset with taupe base

braden-w:
`packages/ui` was sitting on a pre-1.0 shadcn-svelte setup: slate base color, Manrope everywhere, no `style` field, no `iconLibrary`/`menuColor`/`menuAccent` in `components.json`. shadcn-svelte has since shipped the **Luma** style with its own preset picker, and the one we want is `b4IyDhFRSq` — Luma style, taupe base, Lora headings, Inter body, Lucide icons, subtle menu accent. Time to adopt it.

The idiomatic path per the docs is `shadcn-svelte init --overwrite`: it rewrites `components.json`, regenerates the CSS variable block in `app.css`, and reinstalls every registry component in place. That's a "clean break" file-by-file — not a directory nuke — which is both cheaper than rewriting by hand and gives us whatever registry fixes have landed upstream. The downside: it also wipes every Epicenter-local customization the CLI doesn't know about. Finding and surgically re-applying those on top of the new Luma baseline is most of what this PR is.

```bash
bun x shadcn-svelte@latest init --preset b4IyDhFRSq --overwrite --no-deps \
  --css src/app.css --components-alias '#' --lib-alias '#/lib' \
  --utils-alias '#/utils' --hooks-alias '#/hooks' --ui-alias '#'
```

One nice surprise during review: Luma applies `class=\"font-heading\"` directly inside component titles (CardTitle, AlertDialogTitle, AlertTitle, etc.) rather than relying on a global `h1–h6` rule. That's the idiomatic pattern for the Luma style, and the CLI wires it up automatically. Define `--font-heading: \"Lora Variable\", serif` once in `@theme inline` and Lora renders wherever a registry title component is used. No consumer changes, no global CSS reset.

```css
@theme inline {
    --font-sans: \"Inter Variable\", sans-serif;
    --font-heading: \"Lora Variable\", serif;
    /* taupe-derived tokens, chart colors, sidebar tokens… */
}
```

The surgical restorations fall into three buckets. **First, plain exports the CLI clobbered.** `sonner/index.ts` lost its `toast` and `toastOnError` re-exports (used in ten-plus files across apps). `table/index.ts` dropped `SelectAllPopover` and `SortableTableHeader` (both Epicenter-custom, consumed by Whispering's recordings and transformations pages). `utils.ts` and `chart-utils.ts` had their `biome-ignore` comments replaced with `eslint-disable-next-line` — wrong linter for this repo. All restored.

**Second, genuine component behavior that would regress silently.** `button.svelte` lost its custom `tooltip` prop that wraps the button in `<Tooltip.Root>` when set — 67 call sites across Whispering and tab-manager depended on it. `badge.svelte` lost `success` and `id` variants (semantic pills used for status and entity IDs) and three `status.*` variants (more on those below). The whole drawer/dialog z-index story got flattened: everything Luma-generated sits at `z-50`, which works by accident via DOM order but breaks the moment an alert-dialog opens from inside a drawer. Pre-Luma we pinned drawers and regular dialogs to `z-40` so alert-dialogs (at `z-50`) always win. Restored. Same for the vaul-svelte focus-recursion workaround on drawer content (`onOpenAutoFocus={(e) => e.preventDefault()}`), which is still needed because `vaul-svelte@1.0.0-next.7` still depends on `bits-ui@^1` while we're on `^2` — [huntabyte/vaul-svelte#135](https://github.com/huntabyte/vaul-svelte/issues/135). Luma also dropped the scroll wrapper inside drawer/dialog content, so tall content clipped at viewport. Restored.

**Third, behaviors you'd only notice later.** Tooltip provider defaults went from our `300ms`/`150ms` (the desktop-app sweet spot) to Luma's `0ms` — every hover would instantly pop a tooltip across 67 button sites. The Item component lost `relative` on its base (needed for absolute-positioned hover actions), `min-w-0` on title and content (needed for flex text truncation), and Luma's `w-fit` actively prevented shrinking. The Select dropdown lost its `max-w-min` clamp, so long options would expand the dropdown wider than the trigger. Resizable pane-group lost its `gap-2` spacing. All restored, each with a `// Custom:` or `// Custom override:` comment so the next CLI refresh has a prayer of surviving.

Two Luma-design decisions we accepted rather than fought: **destructive variants are now tinted by default** on both button and badge (subtle `bg-destructive/10` instead of solid red), and **RTL logical properties were swept registry-wide** (`ps-/pe-/ms-/me-` → `pl-/pr-/ml-/mr-`). The tinted-destructive shift means we don't need a separate `ghost-destructive` variant anymore — Luma's default `destructive` looks very close to what `ghost-destructive` was. Five call sites across `apps/fuji` and `apps/tab-manager` now use `variant=\"destructive\"` instead. RTL isn't a current concern for these English-primary apps; if it becomes one, it's a separate project.

```typescript
// Before
<Button variant=\"ghost-destructive\" onclick={onRetry}>

// After
<Button variant=\"destructive\" onclick={onRetry}>
```

One sidequest: the `status.completed`/`status.failed`/`status.running` badge variants were a real code smell. They existed to support a dynamic template-literal lookup in `Runs.svelte`:

```svelte
<!-- Before — no compile-time check that status matches a real variant -->
<Badge variant={`status.${run.status}`}>{run.status}</Badge>
```

Deepwiki confirmed what we suspected: shadcn-svelte's idiomatic pattern for status indicators is **composition**, not custom variants — `<Badge variant=\"outline\">` + a Lucide icon + a Tailwind color class. That's what the `DataTableStatus` snippet in the data-table block does. So the three `status.*` variants are out, replaced with a small local `RunStatusBadge` component that maps `'completed' | 'failed' | 'running'` to `CircleCheckIcon` / `CircleXIcon` / `Spinner` plus a green/red/blue text color. Type-safe at the call site, Badge surface aligned with the registry, component lives where it belongs (Whispering, not `@epicenter/ui`).

```svelte
<!-- After — compile-time enforced, idiomatic composition -->
<RunStatusBadge status={run.status} />
```

One known gap worth calling out: tab-manager's `svelte-check` produces nine new \"Cannot find module `#/utils.js`\" errors in the newly-added registry files (`avatar-badge`, `avatar-group`, `avatar-group-count`, plus some item subcomponents). The root cause is pre-existing — `packages/ui/package.json` declares `imports: { \"#*\": ... }` but tab-manager's tsconfig doesn't mirror those as `paths`. It resolves fine at runtime; only svelte-check cares. Fix is a tsconfig paths entry in tab-manager, orthogonal to this PR. Whispering's typecheck is clean against main (27 pre-existing errors in other concerns, zero regressions from this change).

Net change across 249 files: +1292/-834.

### DIFF
--- a/apps/fuji/src/lib/components/EntryEditor.svelte
+++ b/apps/fuji/src/lib/components/EntryEditor.svelte
@@ -7,14 +7,14 @@
 		toDateTimeString,
 	} from '@epicenter/ui/natural-language-date-input';
 	import * as Popover from '@epicenter/ui/popover';
+	import { Spinner } from '@epicenter/ui/spinner';
 	import * as StarRating from '@epicenter/ui/star-rating';
 	import { TimezoneCombobox } from '@epicenter/ui/timezone-combobox';
-	import { Spinner } from '@epicenter/ui/spinner';
+	import type { RichTextHandle } from '@epicenter/workspace';
 	import { DateTimeString } from '@epicenter/workspace';
 	import ArrowLeftIcon from '@lucide/svelte/icons/arrow-left';
 	import Trash2Icon from '@lucide/svelte/icons/trash-2';
 	import { format } from 'date-fns';
-	import type { RichTextHandle } from '@epicenter/workspace';
 	import { goto } from '$app/navigation';
 	import { workspace } from '$lib/client';
 	import type { Entry } from '$lib/workspace';
@@ -76,7 +76,7 @@
 			<span class="text-sm text-muted-foreground">Back to entries</span>
 		</div>
 		<Button
-			variant="ghost-destructive"
+			variant="destructive"
 			size="icon-sm"
 			onclick={() => {
 				confirmationDialog.open({
--- a/apps/fuji/src/routes/trash/+page.svelte
+++ b/apps/fuji/src/routes/trash/+page.svelte
@@ -82,7 +82,7 @@
 										<RotateCcwIcon class="size-4" />
 									</Button>
 									<Button
-										variant="ghost-destructive"
+										variant="destructive"
 										size="icon-sm"
 										title="Delete permanently"
 										onclick={() => {
--- a/apps/tab-manager/src/lib/components/chat/ChatErrorBanner.svelte
+++ b/apps/tab-manager/src/lib/components/chat/ChatErrorBanner.svelte
@@ -25,11 +25,11 @@
 	>
 		<span class="min-w-0 flex-1">{displayMessage}</span>
 		<div class="flex shrink-0 items-center gap-1">
-			<Button variant="ghost-destructive" onclick={onRetry}>
+			<Button variant="destructive" onclick={onRetry}>
 				<RotateCcwIcon class="size-3" />
 				Retry
 			</Button>
-			<Button variant="ghost-destructive" size="icon-xs" onclick={onDismiss}>
+			<Button variant="destructive" size="icon-xs" onclick={onDismiss}>
 				<XIcon class="size-3" />
 			</Button>
 		</div>
--- a/apps/tab-manager/src/lib/components/chat/ConversationPicker.svelte
+++ b/apps/tab-manager/src/lib/components/chat/ConversationPicker.svelte
@@ -121,7 +121,7 @@
 											>{formatRelativeTime(conv.updatedAt)}</span
 										>
 										<Button
-											variant="ghost-destructive"
+											variant="destructive"
 											size="icon-xs"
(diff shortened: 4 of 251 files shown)


braden-w:
Closing this as obsolete.

Current `packages/ui` has moved past this branch: the package docs identify the shipped base as vendored `shadcn-svelte` 1.x on the Vega preset, with Epicenter deltas in `epicenter-overlay.css`. This PR would reorient the package toward Luma, which is now the wrong direction for the current UI system.

Reference: `packages/ui/README.md`.

PR #1709, opened 2026-04-28
fix(whispering): wait for settings before onboarding

samiralibabic:
## Summary
- Wait for the Whispering workspace to load before running the onboarding configuration check
- Reuse one in-flight settings migration promise so startup callers do not race the migration
- Keep the existing transcription configuration checks unchanged

## Why
The onboarding toast currently runs as soon as `AppLayout` mounts. Workspace-backed settings load asynchronously from IndexedDB, so startup can briefly see default settings such as Moonshine with an empty model path before the persisted Whisper C++ selection hydrates. That creates a false persistent “Welcome to Whispering!” setup toast on every launch.

## Testing
- `git diff --check` passes
- `bun run typecheck` in `apps/whispering` still fails on current upstream-level TypeScript/Svelte issues unrelated to this PR

## Changelog
- Fix false onboarding setup toast appearing before saved transcription settings load


samiralibabic:
Closing this as superseded by #1901, which rebases the onboarding/settings readiness fix onto current `main`.

PR #1710, opened 2026-04-28
fix(svelte-utils): refresh tables from result rows

samiralibabic:
Reactive table bindings stopped updating changed rows after the workspace table API moved to the `Result<row | null, error>` shape. `fromTable` was still switching on the older `{ status, row }` response, so async IndexedDB hydration could write valid rows without Svelte state receiving them.

This updates granular table observation to read the current `data` and `error` fields. Valid rows are written into the `SvelteMap`; deleted or invalid rows are removed.

## Changelog
- Fix workspace-backed lists updating after local data finishes loading

## Verification
- Built and installed a local Whispering app with this fix
- Confirmed migrated transformations appear and existing settings still load
- Ran `git diff --check`

`bun run typecheck` for `packages/svelte-utils` was also attempted from the isolated temporary worktree, but it could not resolve workspace dependencies there.


samiralibabic:
Closing this as obsolete. The PR now has an empty diff against `main`, and the current `fromTable` implementation already reads the current `{ data, error }` result shape from `table.get(id)`.

PR #1785, opened 2026-05-20
Collapse sync engine host boundary

braden-w:
This PR keeps the sync engine as a same-process protocol helper and moves host-owned policy back to the route layer.

The room route now resolves a route room for the authenticated subject, producing the internal Durable Object name subject:{subject}:rooms:{room} before the sync engine is called. Auth runs before the sync engine is created. Usage tracking stays with the host route after sync work completes. The sync engine receives only a resolved room name and owns the protocol mechanics: read the request body, reject oversized payloads, call the selected room, and return the binary response plus storage bytes for the host to record.

The boundary is intentional: routes decide who may access a room and how that access is metered; the sync engine decides how to speak the HTTP room sync protocol. WebSocket upgrades and dispatch stay on the room backend adapter because they are pass-through host capabilities, not sync engine behavior.

Checks:
- bun test apps/api/src/sync-engine.test.ts apps/api/src/app.rooms.test.ts
- bun run --filter @epicenter/api typecheck


braden-w:
Hey @braden-w, I am going to close this as superseded rather than trying to rebase it. I walked it against current `origin/main`, not just the stale GitHub diff, and the branch is now 468 commits behind with only two unique commits left: the `SubjectIdentity` to `LocalIdentity` rename and a formatting commit. GitHub also reports it as conflicting, and the local dry merge confirms the conflicts land in places like Drizzle metadata, deleted workspace-sync-doc files, auth barrels, and Svelte session/account UI tests.

The important part is that those conflicts are not ordinary mechanical rebase conflicts. They are mostly evidence that the concepts this branch carried have since been replaced.

The original useful intent was the sync boundary: routes own auth, product policy, metering, and room naming; the sync layer should only handle the Yjs protocol mechanics for an already-authorized room. That idea did survive, but not as this branch implements it. Later work first carried it forward, then deliberately collapsed the separate `sync-engine.ts` and `room-gateway.ts` back into the room route. Current `main` now has the owner-scoped route shape in `packages/server/src/routes/rooms.ts`: `/api/owners/:ownerId/rooms/:roomId`, with `requireOwnership` resolving the owner boundary before the handler builds the Durable Object name.

The Cloud Workspace part is more completely obsolete. The branch added personal Cloud Workspace provisioning, `/api/workspaces`, `/me/apps/:appId/docs/:docId`, and `workspace-sync-doc.ts`. That whole direction was later rejected and reverted: personal workspaces of one and Better Auth organizations as document owners were treated as extra product nouns without real authorization value. Current `main` syncs browser apps through the owner-scoped room path instead, with future sharing left as document ACL work rather than a fake workspace membership check.

The remaining `SubjectIdentity` to `LocalIdentity` rename is also not worth salvaging directly. Main went further and removed that nested local identity shape from the active wire contract. Auth now exposes the flat `userId`, `ownerId`, and `keyring` model, with `OwnerId` as the durable partition key.

So I do not think there is a small follow-up worth stacking on this PR. The useful historical intent is already represented in the newer code and specs: keep host-owned auth and ownership decisions outside the protocol mechanics, avoid workspace/org concepts until they carry real product behavior, and use the owner-scoped room route as the current sync surface. Reopening this branch would mostly reintroduce deleted paths and stale vocabulary.

PR #1827, opened 2026-05-26
refactor(whispering): split operations and Tauri platform services

braden-w:
Whispering had three separate boundaries carrying the same platform idea. The old `query` folder mixed user-facing operations with RPC wrappers, `services/desktop` mixed Tauri-only implementations with web stubs, and runtime checks like `isTauri()` or `__TAURI_INTERNALS__` made every caller rediscover whether a capability was available.

This PR makes those boundaries explicit. User workflows now live under `operations`, RPC wrappers live under `rpc`, service implementations route through `.browser.ts` and `.tauri.ts` files, and `$lib/tauri` is the single Tauri capability namespace. The goal is that Go to Definition lands on the implementation a maintainer needs instead of a defensive runtime proxy or a cross-platform compatibility shell.

```text
Before

query/
  actions.ts              mixed workflows, RPC, delivery, transforms
  desktop/                Tauri-only calls hidden behind runtime checks
services/
  desktop/                shared bucket for unrelated native capabilities
  */index.ts              runtime selection and proxy stubs

After

operations/               app workflows and orchestration
rpc/                      Tauri command wrappers and provider adapters
services/*/
  index.browser.ts        web implementation
  index.tauri.ts          Tauri implementation
tauri.browser.ts          null namespace for web
tauri.tauri.ts            concrete Tauri capabilities
```

The `$lib/tauri` import now has two shapes depending on the call site. Shared code narrows the nullable namespace, while Tauri-only modules use the assertion helper where the filename suffix already proves the runtime.

```ts
import { tauri } from '$lib/tauri';

if (tauri) {
	await tauri.fs.pathToBlob(path);
}
```

```ts
import { requireTauri } from '$lib/tauri';

await requireTauri().fs.pathToBlob(path);
```

That lets the codebase keep the runtime check visible where the runtime is actually ambiguous, and remove the non-null assertions from modules that can only be loaded by Tauri.

## Verification

- `bun run typecheck` in `apps/whispering`, passing with 0 errors and 11 existing Svelte warnings

---
<a href="https://app.blacksmith.sh/EpicenterHQ/codesmith/epicenter/pr/1827"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782380316&installation_id=128463665&pr_number=1827&repository=EpicenterHQ%2Fepicenter&return_to=https%3A%2F%2Fgithub.com%2FEpicenterHQ%2Fepicenter%2Fpull%2F1827&signature=6ef6dcb4f73b3a7dae079acfe24012e9c6a6af9ab5bfa77a849d062d53a14e50"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


braden-w:
Three follow-up commits pushed that round out the Tauri-namespace cleanup:

**1. Extract pure accelerator helpers** (`4fd1cc3`)

The `Accelerator` brand, `isValidElectronAccelerator`, and `pressedKeysToAccelerator` (renamed from `pressedKeysToTauriAccelerator` since the format is Electron's, not Tauri's) lived inside `tauri.globalShortcuts` but had zero Tauri runtime dependency. Moved to `\$lib/utils/accelerator.ts` so they're importable from any platform. The one Tauri-specific call (`os.type()`) got replaced with the existing `IS_MACOS` constant. `ShortcutError` shrinks to just the genuine runtime errors; pure validation variants moved into a new `AcceleratorError` namespace co-located with the helpers.

**2. Prop-drill `tauri` into `syncIconWithRecorderState`** (`0283dd8`)

The helper was called only from inside `if (tauri)` in `AppLayout`, but did its own `tauri?.tray.setIcon(...)` re-narrow because TypeScript can't carry the caller's narrow across a function boundary. Pass `tauri: Tauri` through the signature instead. Eliminates the last `tauri?.` in production code. Same pattern as `GlobalKeyboardShortcutRecorder.svelte`.

**3. Doc updates** (`c838555`)

The `tauri-is-both-the-namespace-and-the-platform-check` article and `services/README.md` both had stale references (`tauri.rpc` sub-namespace, default-import syntax, "browser shim is one line"). Updated to current state and expanded the prop-drilling section to cover both forms:

- **Component-level** (Svelte props): `let { tauri }: { tauri: Tauri } = \$props();`
- **Function-level** (positional parameter): `function syncIconWithRecorderState(tauri: Tauri)`

Plus a new section on `requireTauri()` with the "when to use which" rule:

- Crossing a function or component boundary inside shared code → prop-drill
- Top of a `.tauri.ts` file → `requireTauri()`
- Plain shared code that may or may not be on Tauri → narrow at the call site

After this, production code has zero `tauri!` and zero `tauri?.`. Typecheck passes with 0 errors.

Known follow-up out of scope: `alwaysOnTop.svelte.ts` imports `@tauri-apps/api/window` at top level, which means the Tauri SDK ships in web bundles even though the function is only called inside `if (tauri)`. That's a bundle-size concern, not a narrowing one — separate fix.

PR #1836, opened 2026-05-27
refactor(whispering): clean-break recorder service boundary

braden-w:
The recorder service threaded a `sendStatus` callback through seven function signatures, documented an "emoji and encouraging tone" contract on its type, and let four files emit their own intermediate progress toasts. All of that complexity existed to flash rapid-fire chatter the user couldn't read. Starting a recording used to fire five toasts in under 200ms; cancel fired three, including a "❌ Cancel Failed" that appeared *while* the cleanup continued and was replaced a moment later by "✅ All Done!".

The seam was load-bearing only for the people writing it. From the user's chair, every action already shows a loading toast: the intermediate stages added noise, not signal. From a maintenance chair, every backend re-implemented its own "no device selected" string in its own emoji.

This PR refuses the mid-operation status channel entirely. The service boundary now emits structured facts and nothing else. `operations/recording.ts` is the single owner of every user-facing string in the recording flow.

```
BEFORE                                          AFTER
══════                                          ═════

operations/recording.ts                         operations/recording.ts
   │  loading.update ←──┐                          │  (sole owner of toast copy)
   ▼                    │                          ▼
manualRecorder          │                       manualRecorder
   │  sendStatus ───────┤                          ▼
   ▼                    │                       service.startRecording(params)
RecorderService         │                       session.stop() / .cancel()
   │  sendStatus ───────┤                          │
   ├── navigator        │                          ▼
   │      │  sendStatus ┤                       structured returns:
   │      ▼             │                          { deviceAcquisition }
   │   device-stream    │                          { kind:'artifact'|'blob' }
   │      sendStatus  ──┘                          { status:'cancelled' }
   └── cpal                                        + session.subscribe(state)
          sendStatus
                                                operations/recording.ts maps
9 toast strings across 4 files,                 facts → copy in ONE place
+ a documented "emoji + encouraging
tone" contract in the service type
```

What dies, named explicitly: `UpdateStatusMessageFn` (the callback type plus a docblock prescribing tone), the `sendStatus` parameter on `RecorderService.startRecording` / `RecordingSession.stop` / `RecordingSession.cancel` / `getRecordingStream` / `manualRecorder.startRecording` / `manualRecorder.stopRecording` / `manualRecorder.cancelRecording` / `vadRecorder.startActiveListening`, every `sendStatus({title, description})` call inside the recorder services and `device-stream.ts`, `LoadingHandle.update` (its only consumer was `sendStatus`), and the duplicate fallback-toast authoring that lived in both `cpal.tauri.ts` and `device-stream.ts`. The `buildStartParams` + `resolveServiceForStart` pair in the manual recorder collapses into one `resolveStartDecision(recordingId)` that reads `recording.method` exactly once per start.

Concretely, here is the new shape of a start:

```typescript
// operations/recording.ts
const loading = report.loading({
  title: '🎙️ Preparing to record...',
  description: 'Setting up your recording environment...',
});

const { data: outcome, error } = await manualRecorder.startRecording();
if (error) return loading.reject({ cause: error });

loading.resolve(handleDeviceAcquisitionOutcome(outcome, /* ... */));
```

One loading toast in. One resolution toast out, derived from the structured `DeviceAcquisitionOutcome` the service returned. No callback channel, no copy strings inside the service.

---

**The post-implementation review caught a subtype dance pretending to be symmetry.**

After the sendStatus removal, the `RecorderService` interface still carried a `getActiveRecording()` method that probed Rust for a recording session that outlived a JS reload. The navigator backend implemented it as `Ok(activeSession?.session ?? null)`, where `activeSession` was a closure variable that was always `null` at module init by definition. The navigator probe was dead code wearing the costume of polymorphism.

The first pass at fixing this introduced `CpalRecorderService = RecorderService & { getActiveRecording }` and typed the cpal binding accordingly. That worked, but the `&` was the tell: rehydration isn't "extra recorder service capability." It's a bootstrap concern that exists only because the Rust process keeps the cpal stream alive across JS reloads. Pretending it belongs on the recorder contract makes both shapes harder to reason about.

The greenfield fix is to hoist the probe out of the service entirely and ship it as a free function with the same build-time DI pattern we already use for the service:

```typescript
// types.ts — symmetric again
export type RecorderService = {
  enumerateDevices(): Promise<...>;
  startRecording(params): Promise<...>;
};

export type ProbeActiveCpalRecording = () =>
  Promise<Result<RecordingSession | null, RecorderError>>;

// cpal.tauri.ts — both bound to the same `activeSession` slot
export const CpalRecorderServiceLive: RecorderService = cpal.service;
export const probeActiveCpalRecording: ProbeActiveCpalRecording = cpal.probe;

// index.browser.ts — same null-on-web convention
export const CpalRecorderServiceLive: RecorderService | null = null;
export const probeActiveCpalRecording: ProbeActiveCpalRecording | null = null;
```

The manual-recorder bootstrap reads as one ternary instead of a `Promise.all` that always discarded one half:

```typescript
const bootstrapped = probeActiveCpalRecording
  ? probeActiveCpalRecording().then(({ data }) => { if (data) attach(data); })
  : Promise.resolve();
```

And the dead `if (activeSession) return Ok(activeSession)` shortcut inside the probe is gone too. The probe is only ever called once at module init, when `activeSession` is null by definition; the branch could only have fired if someone called the probe twice, and today nobody does.

---

**While here, two flags from the original review got promoted from "deferred" to "in scope."**

The cpal cancel path used to swallow `commands.cancelRecording()` errors. It would log via `wellcrafted/logger` and return `Ok({ status: 'cancelled' })`, so the user saw "✅ All Done!" even if the Rust side had failed to tear down. That was a behavior parity with the pre-clean-break shape, but it lied to the user. Now the session propagates `Err(RecorderError.CancelFailed)` so `operations/recording.ts` can reject the loading toast and the user actually knows. Teardown still runs unconditionally first so the JS-side state can never wedge in `RECORDING`.

The VAD recorder used to accept an unused `onSpeechRealStart` callback that no external consumer ever passed; the MicVAD event was being forwarded to nothing. Removed.

---

**On the bootstrap pattern itself, since it deserves a moment.**

The race the `bootstrapped` promise guards is real: at module init we kick off an async probe of Rust state, and a user action can fire before that probe resolves. Without the gate, a `cancelRecording()` racing the probe would see a stale `_current === null`, no-op the cancel, and leak the Rust session. The fix is to have every operation `await bootstrapped` before doing anything.

This isn't strange, just necessary — there's no synchronous way to ask Rust about its current state, so we either gate on the probe or pay the latency on every action. Gating once is the right trade.

---

**What we lost, honestly.**

The intermediate "Setting Up your microphone…", "Connecting Device…", "Finding a New Microphone…", "Closing Session…" toasts are gone. They flashed for tens of milliseconds and the user never read them. The browser still shows its own permission prompt UI when the page asks for the microphone, so the "🎯 Almost there!" toast was redundant with the OS-level chrome. The "❌ Cancel Failed" intermediate that flickered between "Cancelling" and "All Done!" was actively misleading — it's gone in the new shape, and a real cancel failure now surfaces as a proper error toast with a humanized title.

Each user action now shows exactly two toasts: a loading state going in, a resolution coming out. Errors still flow through the typed `RecorderError` variants and `loading.reject({ cause })`. The flicker chatter is gone; the actual signal is louder.

If we ever genuinely need mid-flight status back, the right replacement isn't another copy channel — it's a typed phase callback like `onPhase: (phase: 'acquiring-stream' | 'initializing-codec' | 'ready') => void`. Phases, not strings. Keeps copy out of services. We don't need it now.

### DIFF
--- a/apps/whispering/src/lib/operations/recording.ts
+++ b/apps/whispering/src/lib/operations/recording.ts
@@ -56,9 +56,7 @@ export async function startManualRecording() {
 		description: 'Setting up your recording environment...',
 	});
 
-	const { data: outcome, error } = await manualRecorder.startRecording({
-		sendStatus: loading.update,
-	});
+	const { data: outcome, error } = await manualRecorder.startRecording();
 
 	if (error) {
 		loading.reject({ cause: error });
@@ -87,9 +85,7 @@ export async function stopManualRecording() {
 		description: 'Finalizing your audio capture...',
 	});
 
-	const { data: source, error } = await manualRecorder.stopRecording({
-		sendStatus: loading.update,
-	});
+	const { data: source, error } = await manualRecorder.stopRecording();
 
 	if (error) {
 		loading.reject({ cause: error });
@@ -133,9 +129,7 @@ export async function cancelManualRecording() {
 		description: 'Cleaning up recording session...',
 	});
 
-	const { data, error } = await manualRecorder.cancelRecording({
-		sendStatus: loading.update,
-	});
+	const { data, error } = await manualRecorder.cancelRecording();
 
 	if (error) {
 		loading.reject({ cause: error });
@@ -172,7 +166,6 @@ export async function startVadRecording() {
 	});
 
 	const { data: outcome, error } = await vadRecorder.startActiveListening({
-		sendStatus: loading.update,
 		onSpeechStart: () => {
 			report.success({
 				title: '🎙️ Speech started',
--- a/apps/whispering/src/lib/report/index.ts
+++ b/apps/whispering/src/lib/report/index.ts
@@ -45,7 +45,6 @@ export const report = {
 		return {
 			resolve: (r) => emit('success', r, id),
 			reject: (r) => emit('error', r, id),
-			update: (r) => emit('loading', r, id),
 		};
 	},
 };
--- a/apps/whispering/src/lib/report/types.ts
+++ b/apps/whispering/src/lib/report/types.ts
@@ -17,7 +17,6 @@ export type Problem = Notice & { cause: AnyTaggedError };
 export type LoadingHandle = {
 	resolve: (r: Notice) => void;
 	reject: (r: Problem) => void;
-	update: (r: Notice) => void;
 };
 
 export type Level = 'error' | 'success' | 'info' | 'loading';
--- a/apps/whispering/src/lib/services/README.md
+++ b/apps/whispering/src/lib/services/README.md
@@ -250,14 +250,13 @@ export function createManualRecorderService() {
 	return {
 		startRecording: async (
 			recordingSettings,
-			{ sendStatus },
 		): Promise<Result<DeviceAcquisitionOutcome, RecorderError>> => {
 			if (activeRecording) {
 				return RecorderError.AlreadyRecording();
 			}
 
 			const { data: streamResult, error: acquireStreamError } =
-				await getRecordingStream(selectedDeviceId, sendStatus);
+				await getRecordingStream({ selectedDeviceId });
 
 			if (acquireStreamError) {
 				return RecorderError.StreamAcquisition({
--- a/apps/whispering/src/lib/services/device-stream.ts
+++ b/apps/whispering/src/lib/services/device-stream.ts
@@ -9,7 +9,6 @@ import type {
 	Device,
 	DeviceAcquisitionOutcome,
 	DeviceIdentifier,
-	UpdateStatusMessageFn,
 } from '$lib/services/recorder/types';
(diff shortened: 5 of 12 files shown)


braden-w:
Superseded by #1881. This branch predated main's `#platform/recorder` migration and file-inlining rewrites (it still edited the since-deleted `cpal.tauri.ts` / `navigator.ts` / `report/types.ts`), so it could not be merged cleanly. #1881 re-implements the same intent (drop the `sendStatus` status channel, propagate `CancelFailed`, remove the dead VAD `onSpeechRealStart`) cut fresh against today's structure.

PR #1846, opened 2026-05-28
refactor: derive runtime helper return types

braden-w:
Several runtime helper factories were still carrying hand-written return and handle types that duplicated the object literals they return. That made Go-to-Definition land on aliases instead of the implementation and forced public documentation to drift away from the members developers actually use.

This moves those helpers toward derived output types and concrete member annotations where they matter:

```typescript
export function createThing() {
	return {
		open(id: string): ThingHandle {
			// implementation owns the public member shape
		},
	};
}

export type Thing = ReturnType<typeof createThing>;
```

The same cleanup adds a small Svelte result-mutation helper so mutation call sites can share lifecycle handling without each component rebuilding the same pending/error wiring. Dashboard and Honeycrisp consumers use that helper directly, and the shared packages keep tests around both the new mutation utility and the workspace runtime helpers.

---
<a href="https://app.blacksmith.sh/EpicenterHQ/codesmith/epicenter/pr/1846"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782520208&installation_id=128463665&pr_number=1846&repository=EpicenterHQ%2Fepicenter&return_to=https%3A%2F%2Fgithub.com%2FEpicenterHQ%2Fepicenter%2Fpull%2F1846&signature=8f0a790e20dd433fe5d82da8ab59638678bc09c5ff9dd8d42fb10cf2365a04f4"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


braden-w:
Closing this split-review draft because the rebased final branch has already landed on main. This PR was created as a review slice from the pre-landed split, so it is no longer a merge vehicle. Any real leftovers should come back as fresh follow-up PRs from current main.

PR #1877, opened 2026-06-01
fix(server): harden bearer verification and the self-host trust boundary

braden-w:
Bearer-token verification in `require-auth` returned 503 inside Cloudflare because it verified access tokens by fetching the Worker's own `/auth/jwks` endpoint, a same-origin loopback subrequest that does not resolve in the Workers runtime. This branch replaced that with in-process verification against Better Auth's own JWKS projection, and along the way closed three self-host gaps: an AI route that authenticated the caller but never ran the ownership rule, a trusted-origins list that hardcoded Epicenter cloud origins for every deployment (while a self-host did not trust its own origin), and a session cookie pinned to `.epicenter.so`.

This PR is being closed without merging. While it was open, `main` merged #1874 (`refactor/personal-and-shared-wiki`), which renamed the `team` ownership rule to `shared`, moved `apps/team-api` to `apps/self-host`, reframed the self-host deployable as a shared wiki, and independently fixed the AI-route gap this branch also addressed. That left this branch conflicting and built on vocabulary `main` has since removed.

The still-valuable work is being reconciled onto the current `main` in a fresh branch using the `shared`/`self-host` model: the JWKS verification fix (which `main` does not yet have, so the original 503 still reproduces there), the `dispatch_response` recipient binding, and the deployment-supplied trusted-origins and cookie-domain hardening. The AI-ownership change is dropped as already landed. See the closing comment for the reconciliation plan.

### DIFF
--- a/apps/api/worker/index.ts
+++ b/apps/api/worker/index.ts
@@ -63,6 +63,7 @@ mountAssetsApp(app, {
 });
 mountAiApp(app, {
 	auth: requireBearerUser,
+	ownership,
 	policies: [chargeAiCreditsWithAutumn],
 });
 
--- a/apps/api/worker/index.ts
+++ b/apps/api/worker/index.ts
@@ -32,6 +32,7 @@ import {
 	syncAssetStorageWithAutumn,
 } from './billing/policies.js';
 import { mountBillingApi } from './billing/routes.js';
+import { buildEpicenterTrustedOrigins } from './trusted-origins.js';
 
 const ownership = personal();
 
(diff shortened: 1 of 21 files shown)


braden-w:
Closing in favor of a reconciliation onto the post-#1874 `main`.

Triage of this branch's commits against current `main`:

- **JWKS bearer-verify fix** (the original 503): `main` still uses the self-fetching `oauthProviderResourceClient().verifyAccessToken` with `jwksUrl`, so this is still needed and will be re-applied (in-process `verifyJwsAccessToken` fed by `auth.api.getJwks()`).
- **`dispatch_response` recipient binding** and **drop `createOAuthJwksURL`**: still apply, will be re-applied.
- **Trusted-origins + cookie-domain hardening**: still needed, but will be re-expressed against `shared`/`self-host` (the deployment supplies its own trust config; `apps/self-host` trusts only its own origin).
- **AI-route ownership**: dropped, already fixed on `main` as `gate AI behind admission in shared mode`.

A fresh branch off `main` will carry the reconciled set. No production users are affected; nothing here merged.

PR #1883, opened 2026-06-02
fix(server): verify bearer tokens in-process via Better Auth JWKS

braden-w:
## What

Bearer-token auth on the hosted Cloudflare Worker verified each JWT by **fetching the Worker's own `/auth/jwks` over HTTP**. Inside a Worker, a same-zone subrequest to itself does not resolve (loopback restriction), so the JWKS read failed and the request fell back to a retryable **503**. Net effect: bearer auth for external clients (CLI, Tauri, browser extension) was degraded to **503-and-retry** in production.

This switches verification to read the signing keys **in-process** via `auth.api.getJwks()` (Better Auth's own JWKS projection over the same database), removing the HTTP hop entirely.

```
            BEFORE                                   AFTER
  verify    HTTP fetch → own /auth/jwks    →   in-process auth.api.getJwks()
            (loopback fails in CF Worker)        (no network hop)
  401/503   sniff jose ERR_J error codes   →   sentinel matched by identity
  db read   uncaught → 500                 →   try/catch → 503 (infra, not bad token)
```

## Why it is not over-engineered

This change was adversarially grilled twice (the second pass specifically to cut cleverness). What survived and why:

- **In-process `getJwks()` is the fix, not a refactor.** It removes a layer (the `oauthProviderResourceClient` wrapper) rather than adding one. Net: fewer moving parts than before.
- **The 401-vs-503 split is load-bearing, not decoration.** `oauth-resource.ts` maps `ServerError → 503` (client retries) and `InvalidToken → 401` (client discards + refreshes the token). Collapsing them would make a client throw away a good token on a transient server blip.
- **The key-read failure is signalled by a sentinel thrown from the `jwksFetch` callback and matched by identity**, replacing the previous brittle heuristic that string-matched `jose` `ERR_J*` error codes. Traced through better-auth: `getJwks` does not wrap the callback, and `verifyJwsAccessToken` re-throws an `Error` by reference, so identity holds. Throwing (rather than returning `undefined`) also avoids nulling better-auth's module JWKS cache on a transient failure, a subtle bug the flag version had.
- **Keys pass as a callback, not pre-fetched**, so better-auth's module cache serves a known `kid` (and a non-JWT) with **zero database reads**. Intentional perf property.
- **The DB `try/catch` keeps the function's contract**: `resolveRequestOAuthUser` returns OAuth resource errors, so a thrown subject lookup becomes an observable 503 instead of punching through as an uncaught 500.

## Removed (dead after the change)

- `createOAuthJwksURL` helper (its only caller was the loopback)
- `isTokenVerificationError` jose-code heuristic
- the `oauthProviderResourceClient` import in this file (still used in `routes/auth.ts`)

## Notes

Re-implemented fresh on current `main` rather than cherry-picked: the closed #1877 carried this fix bundled with unrelated server work, and `main` has since reworked `require-auth`. This is the single, isolated bearer-verification fix. Verified: `tsc --noEmit` clean; full `packages/server` suite green (99 tests), including a new test asserting keys-unreachable returns 503 (not 401).

---
<a href="https://app.blacksmith.sh/EpicenterHQ/codesmith/epicenter/pr/1883"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782967896&installation_id=128463665&pr_number=1883&repository=EpicenterHQ%2Fepicenter&return_to=https%3A%2F%2Fgithub.com%2FEpicenterHQ%2Fepicenter%2Fpull%2F1883&signature=92a2734b98cc3e9635a7bd892fdcb239c617e4fa591e308d3f372cf0eda00d83"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/server/src/auth/oauth-metadata.test.ts
+++ b/packages/server/src/auth/oauth-metadata.test.ts
@@ -1,7 +1,6 @@
 import { expect, test } from 'bun:test';
 import {
 	createOAuthIssuerURL,
-	createOAuthJwksURL,
 	OAUTH_AUTHORIZATION_SERVER_METADATA_PATH,
 	OAUTH_OPENID_CONFIGURATION_PATH,
 	OAUTH_PROTECTED_RESOURCE_METADATA_PATH,
@@ -10,9 +9,6 @@ import {
 test('OAuth metadata paths follow Better Auth issuer-path layout', () => {
 	expect(createOAuthIssuerURL('https://api.epicenter.so')).toBe(
 		'https://api.epicenter.so/auth',
-	);
-	expect(createOAuthJwksURL('https://api.epicenter.so')).toBe(
-		'https://api.epicenter.so/auth/jwks',
 	);
 	expect(OAUTH_OPENID_CONFIGURATION_PATH).toBe(
 		'/auth/.well-known/openid-configuration',
--- a/packages/server/src/auth/oauth-metadata.ts
+++ b/packages/server/src/auth/oauth-metadata.ts
@@ -11,17 +11,6 @@ export function createOAuthIssuerURL(baseURL: string) {
 	return `${baseURL.replace(/\/+$/, '')}${AUTH_BASE_PATH}`;
 }
 
-/**
- * Return the JWKS URL paired with {@link createOAuthIssuerURL}.
- *
- * Resource servers use this URL with the issuer and audience checks. Keeping it
- * derived from the same base URL prevents metadata routes and token verifiers
- * from drifting apart when the API origin changes.
- */
-export function createOAuthJwksURL(baseURL: string) {
-	return `${createOAuthIssuerURL(baseURL)}/jwks`;
-}
-
 // AUTH_BASE_PATH is hardcoded to '/auth'. The OpenID configuration sits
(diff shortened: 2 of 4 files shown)


braden-w:
Closing as redundant: the in-process JWKS bearer verification and createOAuthJwksURL removal already merged on main via #1880 (commits 904169e4e, fec8c9e70). The only delta was cosmetic (trimmed JSDoc) plus a sentinel-vs-flag refinement for a negligible cache-nulling edge case, not worth a follow-up.

PR #1897, opened 2026-06-04
feat(wiki): runtime-typed pages-and-tags engine over Yjs truth

braden-w:
This branch set out to answer one uncomfortable question: can a person define a *typed* shape at runtime, store it somewhere that merges offline and that agents can mutate without corrupting it, and still walk away with plain markdown they own, all without ever forcing a schema onto anything? The answer turned out to be yes, and this is the ~1600 lines that prove it. It is deliberately small, two tables and a handful of verbs, because the point was never to ship a wiki. The point was to find the smallest engine that makes "store everything, structure it later" actually true.

It opens against `main` as a record, not as a merge. The framing it wears here (a "wiki," one workspace, a knowledge graph) is being replaced by a different one: tenant apps that each materialize into a folder of a vault you already own. But the engine *underneath* that framing is the keeper, and the discussion that reframed it kept re-deriving decisions this branch had already made. So rather than let it rot on a branch or delete it on a whim, this captures what it is and what it taught, so the next pass lifts the proven pieces instead of rediscovering them.

What it is: two ordinary tables:

```
tags  (registry of facets)              pages  (knowledge objects)
  id       youtube_video  (a slug)        id      opaque, == file stem
  name     "YouTube Video"                title
  columns  [ {url}, {duration} ]  <-----  tags    { youtube_video:{url,duration}, idea:{} }
  ...                                     body    (markdown source)

a tag with `columns: []`  is a PLAIN tag      (membership only)
a tag with columns        is a STRUCTURED tag (a typed component -> its own SQLite table)
a page "wears" a tag  iff  its `tags` cell has that key
a page wears a tag at most once; "many of a kind" is an array column, never wearing twice
```

The reason for this shape is:

**1. Truth that merges, output you own.** One authority, two throwaway views.

```
Yjs (Y.Array LWW)   TRUTH      merges offline; agents mutate only via actions
   | one-way
   v
markdown vault      PORTABLE   pages/<id>.md, tags/<id>.md   (read-only)
   | derived
   v
SQLite              QUERY      one table per structured tag   (disposable)

rebuild test:  delete SQLite -> rebuild from Yjs.   delete vault -> rebuild from Yjs.
               delete Yjs    -> nothing rebuilds it.  that is what "truth" means.
```

**2. Types defined at runtime, with no eval and no codegen.** An agent authors a column as a closed `kind`, never a hand-written schema:

```
tags_define({ id: 'youtube_video', columns: [
    { id: 'url',      kind: 'url' },
    { id: 'duration', kind: 'number' } ] })
        |
    buildColumnSchema(kind)        a switch, not an interpreter
        v
    TSchema   (which IS plain JSON Schema)
        |  survives the Yjs / JSON round-trip
        v
    Value.Check still validates it   -> no eval, no parser, no codegen
```

The input layer only accepts the closed `kind` vocabulary, so a stored schema is always a real `column.*` result. Junk can't land.

**3. Changing a type never rewrites your data.** The tag's columns are a lens over the page's stored values, never a gate or a migration (`lens.ts`):

```
page values   { url: "...", duration: 1240, note: "old" }
tag columns   [ url, duration, channel ]
                      |
                viewThroughTag
                      v
    match    url, duration   rendered typed (with a `valid` flag)
    missing  channel         schema has it, page doesn't -> "fill me"
    excess   note            page has it, schema dropped it -> kept verbatim

renaming or reshaping a tag rewrote ZERO pages. the durable data is untouched;
only the rendered view reflects the current schema.
```

**4. The vault is read-only, but the body is writable through verbs.** Agents read `pages/<id>.md` by grepping; they never edit it. Writes go through validated actions: `pages_create`, `pages_assign_tag`, `pages_set_body`, `pages_patch_body`, `tags_define`, `tags_set_column`, `tags_remove_column`. The file is generated; the verbs are the only door.

What already matches where we landed, and what would change:

Already right, and worth lifting forward verbatim:

- tag ids are human slugs (`youtube_video`), so a display rename is metadata-only
- schema lives as *data* (the `tags.columns` cell), authored at runtime
- schema-on-read: loose truth, typed view, lossless type changes
- one-way vault: the markdown is a projection, the actions are the write path

What the reframe changes, and why it's a *next pass* not a fix here:

- the noun: "wiki" becomes the substrate for tenant apps that materialize into a vault's reserved `apps/` folder, browsed alongside your own plain markdown
- `body` is a plain LWW string today; promote it to a per-row `Y.Text` doc once agent / offline body editing is real (the `schema.ts` comment already flags this)
- one workspace today; the tenant model slices into a document per domain (curated notes vs imported messages vs ephemeral tabs) for sync and scale

The implementation works like this:

Start with `schema.ts` (the whole model lives in its doc comment), then `lens.ts` (the one idea that makes loose truth safe), then `markdown.ts` (the projection contract), then `index.ts` (the verbs).

---
<a href="https://app.blacksmith.sh/EpicenterHQ/codesmith/epicenter/pr/1897"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783144757&installation_id=128463665&pr_number=1897&repository=EpicenterHQ%2Fepicenter&return_to=https%3A%2F%2Fgithub.com%2FEpicenterHQ%2Fepicenter%2Fpull%2F1897&signature=bbb332a34525dee5d07a161c808cbdbfd02215015d77681a79cc6cb731d57c87"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/wiki/src/lib/workspace/index.ts
+++ b/apps/wiki/src/lib/workspace/index.ts
@@ -111,15 +111,12 @@ type TagsRegistry = {
  *
  * Typing `#newidea` (assigning a tag that was never defined) upserts
  * `{ id, name: id, columns: [], description: null }`. Capture stays instant;
- * promote later by adding columns. Both the assign path and `markdown_push`
+ * promote later by adding columns. `pages_create` and `pages_assign_tag` both
  * call this, so `page_tags` always resolves to a real row. Invalid slugs are
  * skipped (the caller validates them where the cause is visible); a valid but
- * unwanted slug is acceptable junk, same risk as today's free `string[]` tags.
+ * unwanted slug is acceptable junk, same risk as a free-text tag.
  */
-export function mintMissingTags(
-	tables: TagsRegistry,
-	tagIds: Iterable<string>,
-): void {
+function mintMissingTags(tables: TagsRegistry, tagIds: Iterable<string>): void {
 	const now = DateTimeString.now();
 	for (const tagId of tagIds) {
 		if (!TAG_ID_PATTERN.test(tagId) || tagId === RESERVED_TAG_ID) continue;
--- a/apps/wiki/src/lib/workspace/index.ts
+++ b/apps/wiki/src/lib/workspace/index.ts
@@ -27,6 +27,7 @@ import { defineErrors, type InferErrors } from 'wellcrafted/error';
 import { Ok, type Result } from 'wellcrafted/result';
 import {
 	isTSchemaObject,
+	type Page,
 	type PageId,
 	type PageTypeValues,
 	TYPE_ID_PATTERN,
@@ -54,6 +55,23 @@ export const WikiActionError = defineErrors({
 		message: `Type "${typeId}" column "${columnId}" schema must be a TSchema object (a column.* result)`,
 		typeId,
 		columnId,
+	}),
+	/** A body write named a page id that has no row. */
+	PageNotFound: ({ id }: { id: string }) => ({
+		message: `No page with id "${id}"`,
+		id,
+	}),
+	/** A body_patch anchor (the exact text to replace) was not present in the body. */
+	AnchorNotFound: ({ id, old }: { id: string; old: string }) => ({
+		message: `Anchor not found in page "${id}": the exact text to replace is not present in the body`,
+		id,
+		old,
+	}),
+	/** A body_patch anchor appears more than once, so the target is ambiguous. */
+	AnchorAmbiguous: ({ id, old }: { id: string; old: string }) => ({
+		message: `Anchor is ambiguous in page "${id}": the text to replace appears more than once; include more surrounding context`,
+		id,
+		old,
 	}),
 });
 export type WikiActionError = InferErrors<typeof WikiActionError>;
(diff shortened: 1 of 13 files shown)


braden-w:
Closing unmerged on purpose: this is a reference record, not a merge. The engine is validated and stays linkable here; the next pass lifts the proven pieces (runtime-typed columns, schema-on-read lens, one-way vault, body-write verbs) into the tenant-apps-in-a-vault framing rather than shipping this as a standalone wiki. Branch is kept so the diff stays readable.

PR #1918, opened 2026-06-12
refactor(whispering): move local model lifecycle behind factory boundaries

braden-w:
`LocalModelDownloadCard.svelte` was the whole local model platform in one component: it computed install paths, created directories, streamed HTTP downloads chunk by chunk to disk, validated file sizes, wrote the active model into `deviceConfig`, deleted installs, and fired toasts. `LocalModelSelector.svelte` duplicated another slice of the same filesystem logic for manual imports. None of it was reachable outside those two components, and Whispering's architecture says components don't own business logic.

The lifecycle now flows through two per-model factories, one per layer. The service layer owns disk and network, returns `Result` everywhere, and never touches settings; the operations layer binds that storage handle to the engine's model path setting:

```ts
// services/transcription/local-model-storage.ts
const storage = createModelStorage(model);
await storage.getInstalledPath();          // canonical path when validly installed, else null
await storage.download({ onProgress });    // Result<{ path }, LocalModelStorageError>
await storage.delete();

// operations/local-models.ts
const prebuiltModel = createPrebuiltModel(model);
prebuiltModel.activeModelPath;             // reactive read, tracks deviceConfig in $effect
await prebuiltModel.getStatus();           // 'not-downloaded' | 'ready' | 'active'
await prebuiltModel.downloadAndActivate({ onProgress });
```

The card derives one handle from its prop (`const prebuiltModel = $derived(createPrebuiltModel(model))`) and keeps only its download-progress state machine plus toast copy. The selector keeps the native dialog (it builds titles from the `title` prop) and hands the chosen path to `importModelFile` / `importModelDirectory`.

```txt
constants/local-models.ts            catalog: urls, sizes, directory names
services/fs-paths.ts                 PATHS.MODELS[engine]() model directories
services/transcription/
  local-model-storage.ts             disk: download, verify, import, delete
  providers.ts                       modelPathKey, the one owner of the settings key
operations/local-models.ts           status, activate, download-and-activate, delete
components/settings/*                render state, fire toasts
```

**The engine-to-settings-key mapping now has one owner.** `PROVIDERS` already published `modelPathKey` per local provider, but the root layout re-derived the key with a private template helper and the components spelled it a third way. Operations and the layout now both index `PROVIDERS[engine].modelPathKey`; the template spellings are gone. Similarly, `PATHS.MODELS` is keyed by engine id (`whispercpp`/`parakeet`/`moonshine`) so consumers index it directly instead of switching through a second vocabulary; the on-disk directory names are byte-identical.

**A few behavior fixes fell out of the move.** `download()` returns the path it wrote, so activation is a plain `deviceConfig.set` and can no longer throw from inside a `Result`-typed flow. The double toast on a fresh download ("Model activated" followed by "Model downloaded and activated successfully") is gone; each flow fires exactly one. Opening the settings page no longer `mkdir`s model directories as a side effect of checking status. And a `deviceConfig` change mid-download can no longer clobber the progress UI back to "not-downloaded". Everything durable is untouched: install paths (`models/whisper/{file}`, `models/{parakeet,moonshine}/{dir}`), the `transcription.{engine}.modelPath` keys, and all user-facing copy except one unreachable error toast that the selector's non-empty `models` tuple type now forbids.

Review path: start with `createModelStorage` in `local-model-storage.ts`, then `createPrebuiltModel` in `operations/local-models.ts`, then skim the card as proof the components shrank to rendering. Verified with `bun run --cwd apps/whispering typecheck` (0 errors, no new warnings) and biome. Net: the two components dropped from ~680 lines of mixed concerns to ~420 lines that mostly render.

---
<a href="https://app.blacksmith.sh/EpicenterHQ/codesmith/epicenter/pr/1918"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783865177&installation_id=128463665&pr_number=1918&repository=EpicenterHQ%2Fepicenter&return_to=https%3A%2F%2Fgithub.com%2FEpicenterHQ%2Fepicenter%2Fpull%2F1918&signature=ebc0ac162ed61593146c2499ce9473b2c0d65dcffc42d2630e39451fa0c6dd69"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/whispering/src/lib/components/settings/LocalModelDownloadCard.svelte
+++ b/apps/whispering/src/lib/components/settings/LocalModelDownloadCard.svelte
@@ -7,27 +7,16 @@
 	import CheckIcon from '@lucide/svelte/icons/check';
 	import Download from '@lucide/svelte/icons/download';
 	import X from '@lucide/svelte/icons/x';
-	import { join } from '@tauri-apps/api/path';
-	import {
-		exists,
-		mkdir,
-		remove,
-		stat,
-		writeFile,
-	} from '@tauri-apps/plugin-fs';
-	import { fetch } from '@tauri-apps/plugin-http';
-	import { extractErrorMessage } from 'wellcrafted/error';
-	import { Ok, tryAsync } from 'wellcrafted/result';
 	import { type LocalModelConfig } from '$lib/constants/local-models';
-	import { isModelFileSizeValid } from '$lib/services/transcription/model-file';
-	import { PATHS } from '$lib/services/fs-paths';
-	import { deviceConfig } from '$lib/state/device-config.svelte';
+	import { createPrebuiltModel } from '$lib/operations/local-models';
 
 	let {
 		model,
 	}: {
 		model: LocalModelConfig;
 	} = $props();
+
+	const prebuiltModel = $derived(createPrebuiltModel(model));
 
 	type ModelState =
 		| { type: 'not-downloaded' }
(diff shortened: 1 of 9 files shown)


braden-w:
Two follow-ups from another pass over the final shape:

- `b2dfa56` fixes a race in the card: `refreshStatus` guarded before awaiting the disk check, so a download started during the await could be clobbered by the stale resolved status. The guard now runs after the await.
- `4f5727a` deletes the selector's `fileSelectionMode` prop. It looked like a free choice, but an imported path must match what the engine's preflight accepts, and `PROVIDERS[engine].preflightKind` already declares that fact. The selector derives the mode from the registry, so a call site can no longer pass a mode that preflight would reject.

PR #1922, opened 2026-06-12
feat(whispering): redesign local model settings around recommended defaults

braden-w:
The local-engine section of the transcription settings screen handed the user every decision at once: a two-tab layout per engine, every catalog model as a peer row, and a path picker with step-by-step instructions, all visible before the user had made the only choice that matters (get a working model). This PR rebuilds each engine's card around one obvious happy path: a single prominent action when nothing is installed, the active model's name when one is, and everything else behind progressive disclosure.

The catalog now marks exactly one model per engine as `recommended: true` (Whisper Small for the best size-to-accuracy default, Parakeet TDT 0.6B v3 as the only option, Moonshine Base). Recommended is catalog data, not a user preference; no settings keys changed, and services, operations, and the Rust side are untouched.

Chosen direction (status hero plus one disclosure):

```
+ Whisper Model ------------------------------------+
| Models run on this device...                      |
|                                                   |
|  [empty]   No model installed                     |
|            Download the recommended model...      |
|            [ Download Small (488 MB) ]            |
|            (progress bar while downloading)       |
|                                                   |
|  [active]  | Small          [Active]  (Change) |  |
|            |   488 MB                          |  |
|                                                   |
|  > All models (4)                  [collapsed]    |
|      catalog rows w/ Recommended badge, footer    |
|      > Use your own model          [collapsed]    |
|          help text + path picker                  |
+---------------------------------------------------+
```

Rejected alternatives: a select-like row that expands in place (hides the download CTA behind a click in the empty state, which is exactly the state that needs the strongest affordance), and keeping the flat list with only the recommended row visible plus a "show N more" tail (shows the wrong model at a glance whenever a non-recommended or custom model is active).

The direction matches how comparable local-model apps behave. Ollama's desktop app picks a default model and treats the library as secondary; LM Studio badges a staff pick and puts the size on every download button; superwhisper and MacWhisper recommend one model with its size during onboarding and keep the rest in a secondary list; Handy preselects a recommended default; Wispr Flow goes cloud and deletes the surface entirely. We borrow default-first plus size-on-action and refuse LM Studio's full catalog browser, since our largest catalog is four models. The comparison table now lives in the comparable-apps skill as a worked example for future design passes.

Concretely: `LocalModelSelector` loses its two-tab layout. The empty state is an `Empty.*` block whose primary button reads "Download Small (488 MB)" (sizes stay visible because these are 0.5 to 1.6 GB downloads, possibly on metered connections) and shows a `Progress` bar while downloading; when the recommended model is already on disk but inactive, the button becomes "Activate Small". The active state is an `Item.*` row with the model name, its size (or "Your own model" for manual paths), an Active badge, and a Change button that expands the "All models" `Collapsible`. The catalog rows still use `LocalModelDownloadCard` as the row primitive, now with a Recommended badge. Manual selection lives one level deeper inside "All models" as a "Use your own model" collapsible: a short help paragraph and the same import-on-pick path picker, with the browse button now labeled "Choose file" or "Choose folder" instead of a bare icon.

The page-side snippets shrank to match: `prebuiltFooter` became `catalogFooter`, and the step-by-step `manualInstructions` cards became one-paragraph `manualHelp` descriptions. Moonshine's directory naming requirement (`moonshine-{variant}-{lang}`) survives in its help text since Rust parses the variant out of the directory name; the amber callout styling did not. Moonshine catalog descriptions also dropped their "(~28 MB)" suffixes because size is already its own visible field.

---

**Download state is now shared, not duplicated.** The first cut of this PR gave the recommended-model hero its own private copy of `LocalModelDownloadCard`'s state machine, which meant a download started in one surface was invisible in the other. A follow-up commit lifts that machine into `$lib/state/local-model-downloads.svelte.ts`, a state module keyed by model id (same factory-plus-getter shape as `local-model.svelte.ts`): the module owns disk truth and download progress as `$state`, and the four-state machine (not-downloaded, downloading, ready, active) is computed from those plus the engine's reactive model path. Both components collapsed to a handle lookup and template reads; every `$effect` in this area is gone. Two grounded decisions worth knowing: the state accessor is a plain getter rather than a `$derived` because handles are created lazily from whichever component touches them first, and a derived created inside a component's effect context goes inert when that component is destroyed (Svelte's `derived_inert` warning); and the "All models" expansion stays a component-local `bind:open` boolean, which is the controlled-collapsible pattern shadcn-svelte itself uses in `CodeCollapsibleWrapper`.

Flagged per the brief, controls that survive mainly because they existed before: the "Unload Model When Idle" select still renders below every local engine at top level (it is a real deviceConfig setting, but it predates this pass and was not redesigned), and the X-to-clear button in the path picker remains the only way to deactivate a model without deleting it. One behavior delta from the refactor: disk state is checked once per model handle (first use) and maintained by download and delete, instead of re-checked on every mount; files deleted outside the app mid-session show as installed until the next action.

Verification: `bun run --cwd apps/whispering typecheck` passes with 0 errors (remaining warnings are pre-existing and in untouched files). Not verified in a running Tauri build; the local-engine cards only render under `#if tauri`, so a quick visual pass on the three engines before merge would be worthwhile.

Net effect on the default view per engine: from roughly 12 simultaneous controls down to 3.

### DIFF
--- a/.agents/skills/comparable-apps/SKILL.md
+++ b/.agents/skills/comparable-apps/SKILL.md
@@ -100,6 +100,22 @@ Implication: Epicenter is local-first; the closest references (Obsidian, Logseq)
 
 Asymmetric win surfaced: refusing the feature of "email in chrome" lets the runtime state stop carrying email at all. Email becomes a fetched query at the one surface that wants to display it. See `specs/20260514T210000-profile-as-application-data.md` for the resulting design.
 
+## Worked example: local model pickers (Whispering)
+
+Question: how should a settings screen present downloadable local models (Whisper, Parakeet, Moonshine) when each is a 0.03 to 1.6 GB download?
+
+| App | Category | Picker pattern | Size on action? |
+| --- | --- | --- | --- |
+| Ollama desktop | Local LLM runtime | Picks a default model, downloads on first use; library is secondary | Yes |
+| LM Studio | Local LLM catalog | Full catalog browser with search; "Staff pick" badge | Yes, on every download button |
+| Jan | Local LLM app | Hub with "Recommended" tags | Yes |
+| superwhisper | Local dictation | Onboarding recommends one model with its size; others in a secondary list | Yes |
+| MacWhisper | Local dictation | Model tiers with sizes; recommended default | Yes |
+| Handy | Local dictation (OSS) | Flat model list, recommended default preselected | Yes |
+| Wispr Flow | Dictation (cloud) | No model picker at all; the surface does not exist | n/a |
+
+Implication: local-model apps converge on default-first (one recommended model per engine, size visible on the download action, full catalog secondary). Whispering borrowed default-first plus size-on-action and refused LM Studio's catalog browser, since its largest catalog is four models. Wispr Flow is the refusal boundary: going cloud deletes the surface entirely, which is not Whispering's category. Resulting design: PR #1922 (status hero plus one "All models" disclosure, manual path picker one level deeper).
+
 ## Other questions this lens answers well
 
 ```
--- a/apps/whispering/src/lib/components/settings/LocalModelDownloadCard.svelte
+++ b/apps/whispering/src/lib/components/settings/LocalModelDownloadCard.svelte
@@ -2,13 +2,12 @@
 	import { Badge } from '@epicenter/ui/badge';
 	import { Button } from '@epicenter/ui/button';
 	import { Progress } from '@epicenter/ui/progress';
-	import { toast } from '@epicenter/ui/sonner';
 	import { Spinner } from '@epicenter/ui/spinner';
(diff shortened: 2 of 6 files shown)


braden-w:
Closing rather than rebasing: #1923 merged first and replaced the foundation this branch was built on (settings now store a folder entry name, the file picker import flow and local-preflight are gone). Rebasing would be a rewrite wearing a rebase's clothes.

The intent is preserved for a fresh pass on top of main: the comparable-apps worked example re-lands in #1926 along with a handoff spec (`specs/20260612T164300-local-model-recommended-defaults-rebuild.md`) carrying the design, the grounded decisions, and port notes for the shared download state module. The `recommended: true` catalog flags (d9d7ea9fc) cherry-pick nearly clean and are step one of the rebuild.

PR #1930, opened 2026-06-13
feat(zhongwen): chat over a synced Yjs conversation doc (doc-as-wire)

braden-w:
Zhongwen chat needed the transcript itself to become the transport. The old path split one conversation across a root-doc `chatMessages` table, an SSE response stream, and client runtime state, so persistence, refresh-resume, and multi-device live view all needed extra stitching.

This moves each conversation transcript into its own synced Yjs child doc. The client appends the user message to that doc, then kicks off one server generation; the server joins the same doc as a sync peer and appends assistant tokens back into it.

## Doc-As-Wire Contract

```http
POST /api/ai/chat/doc
Content-Type: application/json

{ "guid": "conversation-child-doc-guid", "generationId": "assistant-message-id" }
```

The POST body carries no message history. The server hydrates the child doc from the room, snapshots the prompt already present there, appends an empty assistant message, flushes token updates into its `Y.Text`, and writes exactly one terminal `finish` value: `completed`, `cancelled`, or `failed`.

```txt
client                                      server
append user message to child doc       ->  POST /api/ai/chat/doc
sync carries the user map                  hydrate room doc
observe child doc                      <-  append assistant map
render partial assistant text          <-  sync token flushes
settle on finish                       <-  sync final finish key
```

`generationId` doubles as the assistant message id, which gives the route a simple idempotency story. A repeated kickoff returns `GenerationAlreadyExists`; a recent unfinished assistant message returns `GenerationInProgress`; a kickoff that arrives before any user message reaches the doc returns `NoUserMessage`. All three are 409s so the client can no-op, wait, or retry without parsing prose.

Stop is just aborting the kickoff fetch. The Worker reads `c.req.raw.signal`, so this PR also enables the `enable_request_signal` compatibility flag in the hosted and self-host wrangler configs.

## Message Ownership

Each child doc owns a `Y.Array('messages')` of append-only `Y.Map`s:

```txt
{
  id: string,
  role: 'user' | 'assistant',
  createdAt: number,
  content: Y.Text,
  finish?: { kind: 'completed' | 'cancelled' | 'failed', ... }
}
```

The client writes user maps. The server generation actor writes assistant maps. Both sides import the same `@epicenter/workspace/ai` helpers, so raw Yjs layout decisions stay behind one module boundary and the two writers cannot drift.

Liveness is derived, not stored. A recent assistant message without `finish` is active; an older unfinished assistant message is an interrupted artifact. The shared active-generation window keeps the client retry UI and the server concurrency guard aligned.

## Zhongwen Shape

Zhongwen now renders chat directly from the conversation child doc. The route owns the durable conversation list and active id, while `ConversationView` owns the per-conversation doc handle, observer, send/stop flow, input draft, and dismissed errors through a keyed component lifecycle.

That removes the old SSE chat client from Zhongwen, including the TanStack chat state, `ui-message.ts`, and the `chatMessages` table. The existing `/api/ai/chat` SSE route remains because tab-manager and opensidian still use it for interactive tool-calling chat, where the client can execute tools and approve work mid-turn.

## Breaking

Zhongwen workspaces no longer define the `chatMessages` table or `ChatMessageId`.

```txt
Before: root doc conversations table + root doc chatMessages table
After:  root doc conversations table + per-conversation child doc transcript
```

Existing `chatMessages` rows are left as orphaned root-doc data. There are no production users to migrate, and keeping a temporary reader would leave two transcript sources alive for no load-bearing reason.

The scope is intentionally bounded:

The unrelated `GOOGLE_CLIENT_ID` var-to-secret change was dropped from this branch after merging current `main`. The only remaining config change is `enable_request_signal`, which belongs to the doc generation abort path.

Conversation delete still leaves the child doc's local storage and relay room behind, and regenerate is still deferred: retry keeps the interrupted partial visible instead of replacing it. Opensidian also needs the same dev `fs.allow` helper cleanup separately.

Spec: `specs/20260612T182359-zhongwen-chat-doc-as-wire.md`.

## Changelog

- Add persistent, synced Zhongwen chat transcripts that survive refreshes and update live across devices
- Add cancellation-aware server generation for Zhongwen chat responses


braden-w (commented):
Reviewed the final cleanup push. The Zhongwen/doc-as-wire code path looks ready from the local and CI quality gates, but this PR should not merge yet because GitHub still reports `mergeStateStatus: BLOCKED`.

Current blockers:
- Deploy Previews fails while building `@epicenter/landing`: Astro cannot render `/whispering` because `bits-ui/dist/bits/accordion/components/accordion.svelte` is being loaded as an external `.svelte` file. The log points to adding the package to the Astro/Vite noExternal path.
- macOS aarch64 preview fails in `@epicenter/whispering` while building `audiopus_sys`: CMake 4 rejects Opus upstream `cmake_minimum_required` compatibility below 3.5; the log suggests `-DCMAKE_POLICY_VERSION_MINIMUM=3.5` as one possible unblock.
- macOS x86_64 preview fails before build because the runner only has `aarch64-apple-darwin` installed and needs `rustup target add x86_64-apple-darwin`.

Quality checks and autofix are green. I would keep this PR open, fix the preview infrastructure blockers separately or confirm they are non-required, then merge after required review plus manual Zhongwen smoke.

braden-w:
Reviewed the final cleanup push. The Zhongwen/doc-as-wire code path looks ready from the local and CI quality gates, but this PR should not merge yet because GitHub still reports `mergeStateStatus: BLOCKED`.

Current blockers:
- Deploy Previews fails while building `@epicenter/landing`: Astro cannot render `/whispering` because `bits-ui/dist/bits/accordion/components/accordion.svelte` is being loaded as an external `.svelte` file. The log points to adding the package to the Astro/Vite noExternal path.
- macOS aarch64 preview fails in `@epicenter/whispering` while building `audiopus_sys`: CMake 4 rejects Opus upstream `cmake_minimum_required` compatibility below 3.5; the log suggests `-DCMAKE_POLICY_VERSION_MINIMUM=3.5` as one possible unblock.
- macOS x86_64 preview fails before build because the runner only has `aarch64-apple-darwin` installed and needs `rustup target add x86_64-apple-darwin`.

Quality checks and autofix are green. I would keep this PR open, fix the preview infrastructure blockers separately or confirm they are non-required, then merge after required review plus manual Zhongwen smoke.

PR #1931, opened 2026-06-13
feat(workspace): one classified scan over four read states

braden-w:
A workspace table read could silently drop your data, in two different ways, and this PR closes both. `getAllValid()` was the default bulk read, and it handed back only the conforming rows: the call path that hides the other states is also the one everyone reached for. Worse, on an encrypted table (which most Epicenter apps run) a row whose key version is missing from the keyring decrypts to nothing, and it was invisible everywhere. `entries()` skipped it, `size` subtracted it, and only a bare counter knew it existed. `count()` returning a number that disagreed with `getAllValid().length` was the desync that started the investigation.

The read surface had grown seven bulk-read methods all walking the same `entries() -> parseRow` boundary (`getAll`, `getAllValid`, `getAllInvalid`, `conformance`, `filter`, plus `find`). This collapses them to one. `scan()` walks every stored entry once and resolves each into exactly one of four mutually exclusive states, returned grouped, with the bucket lengths summing to a separate O(1) `storedCount()`. No read can silently drop data, and the count can never mysteriously disagree with what you can see.

```ts
// before: the valid-only default hides three states; a second scan for the banner
const rows = table.getAllValid();           // drops nonconforming, newer-writer, unreadable
const conformance = table.conformance();    // walks the table again

// after: one walk, the issue buckets ride along at the same call site
const { rows, nonconforming, newerWriter, unreadable } = table.scan();
// storedCount() === rows.length + nonconforming.length
//                 + newerWriter.length + unreadable.length
```

`scan().rows` keeps the conforming payload one property access away. The honest caveat: a mechanical `getAllValid() -> scan().rows` rewrite does not by itself make callers handle the buckets, it relocates the ignore. The win is that the buckets are now present to be acted on, and the materializers do: the markdown and SQLite projections log `scan().nonconforming` rather than silently projecting only `.rows`.

---

**The fourth bucket is the load-bearing one.** Modeling the encrypted blind spot is what makes `storedCount()` reconcile instead of mysteriously exceeding everything you can see. The shared `ObservableKvStore` contract grew `unreadableEntries()` (enumerate the present-but-undecryptable entries, with a reason) and an O(1) `unreadableReason(key)` point probe; plaintext stores return empty. The encrypted store's `size` now includes those entries instead of subtracting them. On an encrypted table with one blob whose key version is absent from the keyring, that row appears in `scan().unreadable`, is counted by `storedCount()`, and the four buckets sum to it. The Fuji repair queue renders it as its own surface, "encrypted with a key this device does not have," which was previously impossible to show.

---

**The classification became a type, not a runtime branch.** The old `version > versions.length` comparison lived buried inside `conformance()`. It is now a first-class `NewerWriter` error emitted in `parseRow` where the version set is in scope, so every consumer switches on it exhaustively instead of re-deriving it. `UnknownVersion` shrank to genuinely corrupt stamps.

---

**The write guard got more complete.** The prior pass refused a stale binary's `set()` over a newer-stamped row (it wins LWW via the monotonic clock and would clobber newer columns on every synced device). That same hazard applies to unreadable rows: `set()` over one used to see `get()` return undefined, treat the slot as empty, and overwrite a row it could not read. `set()`/`bulkSet()`/`clear()` now refuse both states with `TableWriteError` carrying the version or reason, so an import banner can say what it skipped and why. `delete()` stays unguarded, because removing a row you cannot read is a legitimate, key-independent intent.

---

**One reactive binding, one subscription.** `fromTable` and `fromTableConformance` were a redundant double-subscription. They merge: the returned value stays a reactive `SvelteMap` of conforming rows updated granularly per changed id (the hot path, a row changes on every keystroke), now carrying the three issue buckets as debounced properties recomputed from a full `scan()` (the cold path, buckets change rarely). Keeping the map as the primary surface means the dozen-plus existing `fromTable` consumers that only want reactive rows did not change at all.

The decisions are recorded in two ADRs under a new `docs/adr/`, the workspace document README, and the `workspace-api` skill. Every wave is a standalone commit, each green on `bun test` (the workspace package, 507 tests) and `bun run typecheck` across all packages.

Stacked on `lechuguilla-talus` (the prior schema-conformance pass), not `main`, so this diff is exactly the seven read-surface commits. Seven methods became two, one duplicate subscription became zero, and the count finally agrees with what you can see.

---
<a href="https://app.blacksmith.sh/EpicenterHQ/codesmith/epicenter/pr/1931"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783916556&installation_id=128463665&pr_number=1931&repository=EpicenterHQ%2Fepicenter&return_to=https%3A%2F%2Fgithub.com%2FEpicenterHQ%2Fepicenter%2Fpull%2F1931&signature=2b7e44903b0f2a086a49bfe05ce8f0104d12955391a14bb3f5f712f198eae3c8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


braden-w:
Superseded by #1936. Same read-surface work (one classified scan over four read states + the read()/reads()/observe collapse), but extracted cleanly onto `main` and untangled from the landing-page redesign that the stacked base `lechuguilla-talus` carried. #1936 is green on the quality gate (the landing lint error is gone with the landing saga); every workspace commit here is captured there. Closing to consolidate on #1936.

PR #1934, opened 2026-06-13
feat(zhongwen): delete a conversation's local doc and server room

braden-w:
Stacked on #1930 (doc-as-wire). Base is `feat/zhongwen-chat-doc-as-wire`; retarget to `main` once that merges.

The problem this addresses is:

Deleting a Zhongwen conversation only tombstoned the metadata row. Each conversation's transcript is a **separate child Y.Doc**, so its local encrypted IndexedDB database and its server Durable Object room were orphaned **forever**: local storage leaked and the server kept the plaintext transcript indefinitely.

## Design: one tombstone, every layer reacts

```
deleteConversation(id):  conversations.delete(id)   ← CRDT row tombstone, the source of truth (unchanged)
                                 │  (everything below is idempotent, retryable cleanup)
                                 ▼
every device, on a row disappearing (deleter included, no special role):
    whenReleased(handle) → clearLocalStorageForDoc(guid) → DELETE /rooms/:guid
```

Tombstone-the-pointer-first then asynchronously GC the orphan-able resources: the standard soft-delete + GC shape. Cleanup is a pure reactive function of the row tombstone, so it converges across devices with no coordination.

## The server keystone

A Cloudflare Durable Object cannot be truly deleted (the name keeps resolving to a live, empty object), so **drop = empty + tombstone**:

```
destroy():  flag → transactionSync(updateLog.clear(); tombstone.mark()) → deleteAlarm → close sockets(4410)

gated when destroyed:  fetch→410 · webSocketMessage→drop · webSocketClose→no compaction
                       alarm→no compact · sync→discard(no apply) · getDoc→empty snapshot
```

The `webSocketClose`/`alarm` gates close a subtle resurrection path: a post-close compaction would otherwise re-encode the in-memory doc back into the emptied log. The tombstone lives in a separate `room_tombstone` table, read **synchronously** in the DO constructor, so it survives `DELETE FROM updates`, survives eviction, and keeps the no-`await` `blockConcurrencyWhile` invariant.

The `DELETE /api/owners/:ownerId/rooms/:roomId` route is **personal-mode only**: a shared-mode owner partition is a single shared id, so a generic room delete there could tombstone any shared room (including the root doc).

## Commits

1. `feat(workspace)` clearLocalStorageForDoc + cache.whenReleased (generic client primitives)
2. `feat(server)` room destroy + durable tombstone + DELETE route (the keystone)
3. `feat(workspace)` park sync on the room-gone close code (4410)
4. `feat(zhongwen)` reactive cleanup observer + SignedIn.fetch

## Grounding

Designed against DeepWiki (Cloudflare DO deletion semantics, Yjs/y-indexeddb), an independent plan agent, and a Codex adversarial review (which caught the compaction-resurrection path and the shared-mode blast radius). A post-implementation review then fixed a leaky-SQL ownership bug (destroy ran raw `DELETE FROM updates`; now it calls `updateLog.clear()`).

For verification:

- Typecheck: workspace, server, sync, svelte-utils, zhongwen — clean
- `packages/server` tests: 115 pass (5 new destroy tests: no-resurrection sync, close→alarm→compact leaves log empty, 410 upgrade, empty getDoc, reconstructed-room stays destroyed)
- `packages/workspace` tests: 488 pass (whenReleased + clearLocalStorageForDoc)
- Also fixed a latent test-fake bug: `transactionSync` was mocked on `storage.sql` instead of `ctx.storage`

## Known limitations (documented, deferred)

- **Offline sole-device delete** still leaks the server room (local clear succeeds, DELETE never retries). Multi-device/online are covered. Follow-up: a persisted per-guid retry queue.
- **North star**: making "this table field is a child doc" a declared workspace property, so child-doc cleanup is automatic for every app instead of hand-rolled per app.

Note: this branch is one `chore: apply formatting` commit behind the base; the merge is trivial.

---
<a href="https://app.blacksmith.sh/EpicenterHQ/codesmith/epicenter/pr/1934"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783919985&installation_id=128463665&pr_number=1934&repository=EpicenterHQ%2Fepicenter&return_to=https%3A%2F%2Fgithub.com%2FEpicenterHQ%2Fepicenter%2Fpull%2F1934&signature=3b27817e2f864b49e39d2c4ad4827a66e8eecc523257dac450f54d977173177a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/zhongwen/zhongwen.browser.ts
+++ b/apps/zhongwen/zhongwen.browser.ts
@@ -15,8 +15,10 @@
  */
 
 import type { SignedIn } from '@epicenter/svelte';
+import { ROOM_ROUTE } from '@epicenter/sync';
 import {
 	attachLocalStorage,
+	clearLocalStorageForDoc,
 	createDisposableCache,
 	type DeviceId,
 	defineWorkspace,
@@ -24,12 +26,35 @@ import {
 	roomWsUrl,
 	wipeLocalStorage,
 } from '@epicenter/workspace';
+import { defineErrors } from 'wellcrafted/error';
+import { createLogger } from 'wellcrafted/logger';
 import * as Y from 'yjs';
 import {
 	type ConversationId,
 	createZhongwen,
 	zhongwenConversationDocGuid,
 } from './zhongwen';
+
+const log = createLogger('zhongwen/browser');
+
+const ZhongwenCleanupError = defineErrors({
+	/**
+	 * A deleted conversation's child doc could not be fully cleaned up. The local
+	 * clear or the server room DELETE failed (e.g. offline). The conversation is
+	 * already gone from the table; this leaves orphaned storage to reclaim later.
+	 */
+	ConversationCleanupFailed: ({
+		cause,
+		conversationId,
+	}: {
+		cause: unknown;
+		conversationId: ConversationId;
+	}) => ({
+		message: `[zhongwen] failed to clean up deleted conversation ${conversationId}`,
+		cause,
+		conversationId,
+	}),
+});
 
 export function openZhongwenBrowser({
 	signedIn,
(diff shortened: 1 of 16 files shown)


braden-w:
Closing: not building server/client-side deletion reclaim for now (greenfield review).

The only driver was reclaiming a deleted conversation's orphaned storage, but the cost is decision-irrelevant. A SQLite-backed Durable Object bills only stored bytes when idle (~12 KB floor + a tens-of-KB transcript, on the order of $0.0001/conversation/month), and the parent-doc CRDT row tombstone (the source of truth) already removes the conversation everywhere it is visible. With gc: true, that tombstone's own footprint is capped to tiny mergeable GC structs.

Room.destroy() was also built on a false premise ('a Durable Object cannot be truly deleted'). Per Cloudflare docs, ctx.storage.deleteAll() + deleteAlarm() truly deletes a DO (empty-on-evict means it ceases to exist), and writing a room_tombstone row is counterproductive: it keeps the DO permanently non-empty and therefore permanently billable. So the tombstone table, the room-gone close code, and EMPTY_DOC_SNAPSHOT (which also crashed the worker on boot via a module-scope Y.Doc) were paying permanent maintenance and storage cost to save fractions of a cent.

If we ever want true purge-on-delete, it is a ~3-line destroy() (deleteAll() + deleteAlarm()) plus a personal-mode DELETE route, authored on demand. Deferring until storage is a real line item or we make an explicit purge promise.

PR #1952, opened 2026-06-13
feat(transcription): improve setup defaults

braden-w:
Fresh transcription setup now matches the runtime the user is actually in. Desktop workspaces default to the recommended local Parakeet path, while browser workspaces default to OpenAI so web users start on a configurable provider instead of a desktop-only local model.

The settings page also handles synced local selections on web more clearly: if a desktop-only provider is selected, web shows an explicit warning instead of an empty local-model setup area. Internally, upload adapters now use `spokenLanguage` for the source-language hint, keeping the old persisted `transcription.language` key stable while removing the misleading `outputLanguage` vocabulary.

This also adds a small `/setup` route for onboarding into the same runtime setup surface used by transcription settings. The route points incomplete users at the concrete missing setup step and sends them back home once the runtime is ready.

## Changelog

- Fresh desktop transcription setup defaults to Parakeet.
- Fresh web transcription setup defaults to OpenAI.
- Web settings warn when a synced local transcription service cannot run in the browser.
- Cloud and self-hosted transcription adapters now use `spokenLanguage` for the source-language hint.
- A new setup route reuses the transcription runtime settings and is linked from onboarding.


braden-w:
Superseded by #1956, which cherry-picked only the Whispering commits onto a clean branch and has been merged.

PR #1957, opened 2026-06-13
refactor(workspace): one epicenter.config.ts declares a single Mount

braden-w:
An `epicenter.config.ts` default-exported a `Mount[]`, but every config in the repo was an array of one, and the loader carried a branch whose only job was to tell authors to wrap their single mount back into an array. That is an unused multi-mount capability taxing the shape people instinctively reach for. This refuses it: one Epicenter folder is one app is one mount, and the config declares it directly.

```ts
// before
export default [fuji()];

// after
export default fuji();
```

The layout half (`apps/<name>/` projection becoming `<root>/<name>/`) already landed on main; this is the cardinality half, plus the two collapses it unlocks.

This change:

**1. Cardinality.** `loadProjectConfig` returns `Result<Mount, _>`, the validation inverts (a bare mount is accepted, an array is rejected with a pointer to export it directly), and `openProject` opens the one declared mount and returns a single `StartedMount`. The parallel-open and dispose-siblings machinery that only existed for multi-mount configs is gone; `runUp` wraps the one started mount into the daemon's served set, which still routes IPC by name.

**2. One owner per validation concern.** The mount name was format-checked twice on the same string, once at open time and again at bind time. Split by what each layer can actually know:

```
format   (static, per-name)   ->  the loader      (earliest point; the only place
                                                    the error can point at the file)
uniqueness (cross-mount)      ->  the daemon server (the only place a served set
                                                    can collide; one config can't
                                                    collide with itself)
```

Both reuse one `isValidMountName`, so the regex has a single home. The redundant open-time check and the now-dead `MountRejected` error variant are deleted.

**3. One module for the mount story.** The "what is a mount and how does it come online" story was smeared across three directories, with the `Mount` type living apart from its runtime witness (`isMount`) and its name rule.

```
before                                  after
config/                                 mount/
  load-project-config.ts                  contract.ts     Mount, MountContext, defineMount,
  project-config-source.ts                                DaemonRuntime, StartedMount, isMount,
workspace-apps/                                           isValidMountName, validateMountNames
  open-project.ts                         config-source.ts  load-config.ts
  errors.ts                               open.ts  open-errors.ts  auth-client.ts
  auth-client.ts                        daemon/         <- now only the running socket process
daemon/
  define-mount.ts                       (config/ and workspace-apps/ deleted)
  mount-validation.ts
  types.ts (Mount runtime types)
```

`mount/contract.ts` is the single, drift-proof definition of a valid mount: the type, the runtime it produces, the guard, and the name rule, all browser-safe. `daemon/types.ts` keeps only the narrowed served view (`DaemonServedMount`), so the `mount -> daemon` type edge disappears and `mount/` is a clean leaf the daemon builds on.

## Ownership decision, and what this refuses

The driving question was "is there a fat layer to delete now that config is singular?" The answer, after grilling it from several angles, is **no**, and that is a deliberate result. These were considered and refused:

- **Merge `openProject` into `runUp`.** Kept. It is the workspace/CLI package seam that keeps `MountContext` and keyring assembly out of the CLI, and it is independently tested.
- **Kill the daemon's served-set / route-by-name.** Kept. The `<mount>.` action prefix, the set, and the lookup are one abstraction; the mount name is the action namespace regardless of how many mounts a daemon serves.
- **Singularize `validateMountNames`.** Kept plural. The daemon server owns cross-mount uniqueness, which only the assembled set can know.

This is a pure relocation at the package boundary: every barrel re-exports the same symbols from the new paths, so `@epicenter/workspace`, `/node`, and `/daemon` are byte-identical to consumers and the CLI and every mount author (fuji, honeycrisp, opensidian, zhongwen, tab-manager) are untouched.

One behavior change worth calling out for anyone writing a config: a freshly `init`-ed folder no longer runs until you add your mount, because a zero-mount config is no longer meaningful. The scaffold is now a comment-only template that says so.

Deliberately out of scope: `defineMount` is still published from three barrels (root, `/daemon`, `/node`). Collapsing that to one canonical home changes where authors import it from, so it is a separate, consumer-facing change.

---
<a href="https://app.blacksmith.sh/EpicenterHQ/codesmith/epicenter/pr/1957"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783974649&installation_id=128463665&pr_number=1957&repository=EpicenterHQ%2Fepicenter&return_to=https%3A%2F%2Fgithub.com%2FEpicenterHQ%2Fepicenter%2Fpull%2F1957&signature=1493e3257a9623383e86ac9a0d32dd842a8a46df9f53ce2c5cb3838dcc9fdf38"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


braden-w:
Closing: the headline change here (singular `export default fuji();`) already landed on main through other work, and the remaining module reorg (`config/` + `workspace-apps/` -> `mount/`) diverged from how main actually evolved (main kept and extended `config/`), which is why it's a 25+ conflict rebuild. Nothing unique is lost. If the `mount/` consolidation is still wanted, it's cleaner to redo it fresh against current main.

PR #1970, opened 2026-06-14
fix(whispering): reload the local model after re-download

braden-w:
Re-downloading a local transcription model (after deleting it, or after a corrupted download) left two stale surfaces until you manually re-selected the model.

The resident model is pushed to Rust by the `$effect` in `(app)/+layout.svelte`, which keys off the selected model *name*. A delete + re-download keeps the same name, and Svelte's `SvelteMap.set(key, sameValue)` does not notify on an unchanged value, so the effect never re-fired and Rust was never told to drop and reload the model now sitting at the same path. Re-selecting forced a name change, which is why that worked around it. This adds a tiny bump-only reactive signal (`transcription-reload`) that the download-completion path bumps and the layout effect reads, so a same-name re-download re-fires the config push. The push stays a single site; the signal only re-triggers it.

Separately, the "Downloaded" badge in the model dropdown could lag because `LocalModelSelector.refreshEntries` fired the per-model disk-stat refresh without awaiting it, racing the render. It now awaits them.

## Changelog

- Re-downloading a local transcription model now takes effect immediately, with no need to un-select and re-select it.
- The "Downloaded" badge updates reliably after a model finishes downloading.

---
<a href="https://app.blacksmith.sh/EpicenterHQ/codesmith/epicenter/pr/1970"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784012445&installation_id=128463665&pr_number=1970&repository=EpicenterHQ%2Fepicenter&return_to=https%3A%2F%2Fgithub.com%2FEpicenterHQ%2Fepicenter%2Fpull%2F1970&signature=150994bf47ea90d7e16a02e28ecf96adae530d81b0266abeb149ccf0059d4504"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/whispering/src/lib/components/settings/LocalModelSelector.svelte
+++ b/apps/whispering/src/lib/components/settings/LocalModelSelector.svelte
@@ -121,10 +121,10 @@
 		if (!tauri) return;
 		entries = await listModelEntries(engine);
 		// The folder is user-editable truth, so the catalog handles re-check
-		// disk on the same signal that rescans the folder.
-		for (const model of models) {
-			localModelDownloads.get(model).refresh();
-		}
+		// disk on the same signal that rescans the folder. Await the disk-stat
+		// so `isInstalled` (and the "Downloaded" badge it drives) is settled
+		// before the listing does, instead of racing the next render.
+		await Promise.all(models.map((model) => localModelDownloads.get(model).refresh()));
 		// A user who already brought their own model gets the list, not a
 		// download pitch: when nothing is active and the first scan finds
 		// custom entries, start with the list open instead of the hero.
--- a/apps/whispering/src/lib/state/local-model-downloads.svelte.ts
+++ b/apps/whispering/src/lib/state/local-model-downloads.svelte.ts
@@ -28,6 +28,7 @@ import {
 	deleteModelEntry,
 	type LocalModelFolderError,
 } from '$lib/services/transcription/local-model-folder';
+import { transcriptionReload } from '$lib/state/transcription-reload.svelte';
 
 type ModelDownloadState =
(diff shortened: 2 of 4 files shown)


braden-w:
Closing as superseded by #1974.

The headline goal here (reload a local model after a delete + re-download under the same name) is fixed at the source in #1974 via disk-identity revalidation: the engine cache fingerprints the bytes it loaded from (length + aggregated mtime, following symlinks, recursive for directory models) and reloads when they change, regardless of whether the FE re-pushes config.

That makes the `transcriptionReload` signal + layout dependency in this PR unnecessary. It also would not have fixed the bug on its own: the old `set_transcription_config` routes through `should_preload`, which returns false when `(engine, model_name)` is unchanged, exactly the re-download-same-name case, so the re-push was discarded by the backend before it could evict anything.

The one orthogonal good fix here, awaiting the per-model disk refresh so the "Downloaded" badge settles, was salvaged verbatim into #1975 (commit 0c2acbfb5). Nothing is lost by closing.

PR #1986, opened 2026-06-14
docs(spec): record Wave 3 decisions, the Wave 4 publish blocker, and mark Waves 1-2 landed

braden-w:
Follow-up to #1980 (merged), which landed Waves 1 and 2 of the app-folder-as-root topology. Documentation only: updates `specs/20260614T120000-app-folder-as-root-and-jsrepo-blocks.md`.

## Wave 1 / Wave 2

Marked `[landed in #1980]`.

## Wave 3 (the `--all` fan-out): decisions recorded

- **Process model:** a foreground supervisor that spawns one child `epicenter daemon up -C <root>` per discovered root. Child processes (not N mounts in one process) preserve physical fault isolation and reuse the single-root lifecycle unchanged.
- **Partial failure:** start-healthy-and-report.
- **Discovery:** the `--all` scan is the only scan-down path; stops at the first config per branch; skips `.epicenter`, `node_modules`, `.git`.
- **Naming:** `daemon up --all` as a pure dispatch to a separate `runSupervisor()`, never generalizing `runUp`.
- Resolves the decided parts of **Open Question 2**.
- Records the asymmetry motivating Wave 3: `daemon ps` and `daemon down --all` are already container-global; only `up` is single-root.

Wave 3 stays **deferred**: real demand is downstream of Wave 4.

## Wave 4 (jsrepo distribution): BLOCKED on the npm publish pipeline

An audit found `jsrepo add fuji` cannot work end to end today because the `@epicenter/*` framework closure is uninstallable from npm:

- **Incomplete publish set:** `@epicenter/workspace` depends on `encryption`, `field`, `identity`, `util` (all 404 on npm: `private: true` or never changeset-published). Only `sync`/`workspace` ship.
- **Unresolved protocols:** published manifests ship `workspace:*` and `catalog:` verbatim. The release path is `changeset version && changeset publish` (changesets over npm), and `catalog:` is a bun-only protocol nothing in that path rewrites.

Wave 4 is rewritten as a four-phase plan gated on **Phase 0** (fix the publish pipeline: publish-or-bundle the closure, resolve `catalog:` at publish). Also records that the real block candidate is `examples/fuji` (currently stale, needs a Wave 1 refresh), not the SvelteKit `apps/fuji`.

### DIFF
--- a/specs/20260614T120000-app-folder-as-root-and-jsrepo-blocks.md
+++ b/specs/20260614T120000-app-folder-as-root-and-jsrepo-blocks.md
@@ -178,7 +178,7 @@ DAEMON PATHS  already per-root (NO CHANGE)
 
 Sequenced so each wave is independently shippable and reviewable.
 
-### Wave 1: Projection convention + gitignore (behavior-preserving prep)
+### Wave 1: Projection convention + gitignore (behavior-preserving prep) [landed in #1980]
 
 1. Change the projection to root-relative. Mount factories pass `dir:
    epicenterRoot` (or a chosen content dir), letting the exporter's per-table
@@ -192,7 +192,7 @@ Sequenced so each wave is independently shippable and reviewable.
    Keep `.epicenter/.gitignore = *`. The jsrepo block MAY ship a tracked
    `.gitignore` for belt-and-suspenders, but vendored source is tracked by default.
 
-### Wave 2: Cardinality collapse (the #1957 idea, rebuilt against current main)
+### Wave 2: Cardinality collapse (the #1957 idea, rebuilt against current main) [landed in #1980]
 
 4. `loadEpicenterConfig` returns `Result<Mount>`. Accept a bare `Mount`, reject a
    `Mount[]` with a pointer to unwrap it. Move format validation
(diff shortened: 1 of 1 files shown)


braden-w:
Closing as obsolete. Every load-bearing decision recorded here already shipped on main in executable form, so the doc now frames solved work as open:

- Waves 1–2: `examples/fuji/epicenter.config.ts` is singular `export default fuji()`; `load-epicenter-config.ts` returns `Result<Mount>` and rejects arrays.
- DEFECT 1 (package closure): `packages/encryption` and `packages/util` deleted; `field`/`identity`/`sync`/`workspace` published at 0.3.0 with `files` allowlists.
- DEFECT 2 + publish policy: `scripts/publish-packages.ts` uses `bun publish` to resolve `catalog:`/`workspace:*`; `.changeset/config.json` `fixed` group includes `field`/`identity`; README documents the `release:dry-run` gate.

Only unlanded content is the Wave 3 fan-out supervisor design (`up --all`). If worth preserving, re-file a small PR adding only that section rather than the 181-line block re-litigating settled publish decisions.

PR #2008, opened 2026-06-15
refactor(workspace): brand content-doc guids and own child-doc construction

braden-w:
Two commits, each reviewable on its own. The guid string is unchanged from main (4-part `workspaceId.collection.rowId.field`), so no stored data moves.

## 1. Brand the guid and unify its validator (`25dd2be`)

- **`DocGuid` brand.** `docGuid` returned `Guid`, the same brand as `generateGuid()` (a random token). A composed dotted guid and a random token obey different grammars, so they now carry distinct brands.
- **One grammar, one validator.** The server hand-copied the segment grammar as a local regex (`DOC_GUID_REGEX`). That is now `isDocGuid`, derived from the same `isSafeSegment` predicate `docGuid` mints with, so a validator and a minter cannot drift.
- **`field` kept.** An earlier pass dropped it to a 3-part guid on the theory that every row owns one child doc. It is what lets a row own more than one sibling child doc (e.g. an article with a `code` doc and a `preview` doc, see `docs/articles/updated-at-sentinel-for-external-yjs-docs.md`) without the two colliding. Dropping it would rest injectivity on an unenforced convention, and a second child doc would merge silently into the first.

## 2. Make the brand load-bearing via `createContentDoc` (`b56d5fa`)

After commit 1 the brand was produce-only: `new Y.Doc({ guid })` takes a plain `string`, so nothing required a `DocGuid` and the per-app wrappers widened it back to `string`. A brand no consumer demands is decorative.

- **`createContentDoc(guid: DocGuid): Y.Doc`** is now the one place a child content document is allocated. Requiring a `DocGuid` turns "passed a raw row id / random guid to child-doc construction" into a compile error instead of a silent collision into the wrong doc. It also owns the one shared construction policy (`gc: true`); runtime concerns (persistence, sync, codecs, caching) stay at call sites.
- **Threaded through every content-doc site** (all were a uniform `new Y.Doc({ guid, gc: true })`): fuji, honeycrisp, opensidian, zhongwen, the skills package (browser + node), opensidian skill-state, the filesystem test, and the one-shot `readRoomOverHttp`. The six wrappers now return `DocGuid`, so the brand survives end to end to the constructor.

## Surface

- `packages/workspace`: `DocGuid` brand, `isDocGuid`, `isSafeSegment`, `createContentDoc`; `docGuid` returns `DocGuid`; `readRoomOverHttp` takes a `DocGuid`.
- `packages/filesystem`, `packages/skills`: content-doc wrappers return `DocGuid`; openers use `createContentDoc`.
- `packages/server`: doc-as-wire guid validated with `isDocGuid` instead of `DOC_GUID_REGEX`.
- `apps/*` (fuji, honeycrisp, opensidian, zhongwen, skills): construct content docs with `createContentDoc`.

### DIFF
--- a/apps/fuji/src/lib/workspace/browser.ts
+++ b/apps/fuji/src/lib/workspace/browser.ts
@@ -21,6 +21,7 @@ import {
 	type ActionRegistry,
 	attachLocalStorage,
 	attachRichText,
+	createContentDoc,
 	createDisposableCache,
 	DateTimeString,
 	type DeviceId,
@@ -30,7 +31,7 @@ import {
 	roomWsUrl,
 	wipeLocalStorage,
 } from '@epicenter/workspace';
-import * as Y from 'yjs';
+import type * as Y from 'yjs';
 import { createFuji, type EntryId, entryContentDocGuid } from './index';
 
 export function openFujiBrowser({
@@ -78,7 +79,7 @@ export function openFujiBrowser({
 	});
 
 	const entryBodies = createDisposableCache((id: EntryId) => {
-		const ydoc = new Y.Doc({ guid: entryContentDocGuid(id), gc: true });
+		const ydoc = createContentDoc(entryContentDocGuid(id));
 		// Sync is opened for its side effect; the collaboration handle is
 		// orphaned on purpose, teardown cascades from `ydoc.destroy()` below.
 		const { idb: bodyIdb } = wire(ydoc, { actions: {} });
--- a/apps/fuji/src/lib/workspace/index.ts
+++ b/apps/fuji/src/lib/workspace/index.ts
@@ -32,6 +32,7 @@ import { field } from '@epicenter/field';
 import {
 	createWorkspace,
 	DateTimeString,
+	type DocGuid,
 	defineActions,
 	defineMutation,
 	defineQuery,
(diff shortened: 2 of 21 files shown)


braden-w:
Closing as superseded. Main reached the same destination (workspace owns child-doc construction) by a cleaner route: table-declared `.docs({...})` from the trusted-relay child-docs line (#2020–#2027). Concretely this PR now patches code that no longer exists on main:

- `createContentDoc` seam never landed; inline `new Y.Doc({guid:…})` construction is gone, replaced by `.docs(...)`. `entryContentDocGuid` was deleted.
- The guid-helper files it edits (`filesystem/file-content-docs.ts`, `skills/reference-content-docs.ts`, `skill-instructions-docs.ts`) are deleted on main.
- It re-exports `docGuid` + adds `DocGuid`, contradicting main's deliberate decision (`workspace/src/index.ts`) to keep `docGuid` internal and expose only `tables.<table>.docs.<field>.guid(rowId)`.

Only surviving idea: `server/src/routes/ai.ts` still validates the content-doc guid with a hand-copied `DOC_GUID_REGEX`. If worth fixing, that's a tiny standalone PR (add `isDocGuid` to `doc-guid.ts`, export only it, swap the regex in `ai.ts`) — not this child-doc refactor.

PR #2050, opened 2026-06-16
refactor(whispering): source shortcut reset defaults from the schema

braden-w:
Resetting an in-app shortcut to its default read from `DEFAULT_LOCAL_SHORTCUTS`, a hand-maintained map in `operations/shortcuts.ts`, even though every one of those defaults already lived in the synced workspace schema (`definition.ts`, via `defineKv`). Two copies of the same `p`/`space`/`c`/`v`/`t`/`r` table, kept in sync by hand. The global reset had the same shape against `DEFAULT_GLOBAL_BINDINGS`.

This also repays an invariant that #2048 spent. Gating the transformation picker behind the `#platform/commands` seam made `Command['id']` platform-varying (the picker is absent from `commands` on web), which forced `DEFAULT_LOCAL_SHORTCUTS` to loosen from `Record<Command['id'], …>` to `Record<string, …>` and drop its "every command has a default" check. Sourcing from the schema removes the map entirely, so that check is no longer something to lose: the schema is the exhaustive source, validated by `defineKv`.

## Old shape

```
operations/shortcuts.ts
  DEFAULT_LOCAL_SHORTCUTS = { pushToTalk: 'p', … }   // parallel copy of the schema
  resetLocalShortcuts:  settings.set(key, DEFAULT_LOCAL_SHORTCUTS[id] ?? null)
  resetGlobalShortcuts: deviceConfig.set(key, DEFAULT_GLOBAL_BINDINGS[id] ?? null)
```

## New shape

```
resetLocalShortcuts:  settings.set(key, settings.getDefault(key))        // synced schema
resetGlobalShortcuts: deviceConfig.set(key, deviceConfig.getDefault(key)) // device-config schema
```

A shortcut's default now lives in exactly one place per surface. Locals resolve through `settings.getDefault` (the synced workspace definition, the same source the settings UI already shows); globals resolve through `deviceConfig.getDefault` (the device-config definition). The local/global asymmetry is preserved on purpose: globals are device-local and must not come from the synced schema, so each reset routes through its own definition rather than a shared map.

`DEFAULT_GLOBAL_BINDINGS` stays, now private, in its narrowed role of seeding the `defineEntry` defaults; it is no longer exported (its only external consumer, the old `register-commands`, is gone) and no longer read by the reset path. The `?? null` fallbacks are dropped because `getDefault` always returns a value for a valid key.

---
<a href="https://app.blacksmith.sh/EpicenterHQ/codesmith/epicenter/pr/2050"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784219060&installation_id=128463665&pr_number=2050&repository=EpicenterHQ%2Fepicenter&return_to=https%3A%2F%2Fgithub.com%2FEpicenterHQ%2Fepicenter%2Fpull%2F2050&signature=d6d20563c469c227b41e42a16d31c91e49916ddaf946e0561f1d7e1522ba3b9f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/whispering/src/lib/operations/shortcuts.ts
+++ b/apps/whispering/src/lib/operations/shortcuts.ts
@@ -10,10 +10,7 @@ import {
 	type CommandId,
 	shortcutStringToArray,
 } from '$lib/services/local-shortcut-manager';
-import {
-	DEFAULT_GLOBAL_BINDINGS,
-	deviceConfig,
-} from '$lib/state/device-config.svelte';
+import { deviceConfig } from '$lib/state/device-config.svelte';
 import { settings } from '$lib/state/settings.svelte';
 import type { CommandBinding, KeyBinding } from '$lib/tauri/commands';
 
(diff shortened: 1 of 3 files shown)


braden-w:
Closing as a duplicate of #2051. Both PRs are the same collapse (source shortcut reset defaults from the schema) across the same three files. #2051 is rebased onto current main and is the one moving forward; it also extends the collapse to the display path. Deleting the stale `whispering-defaults-schema-source` branch this PR was built from.

PR #2056, opened 2026-06-16
feat(whispering): add encrypted secret vault foundation

braden-w:
Whispering provider keys need a path out of raw device config before they can sync across devices. This PR builds that path in layers: an encrypted LWW key-value primitive, passphrase-derived vault keys, live relay sync coverage, and a Whispering `secrets` facade that gives consumers one read contract: `available`, `locked`, or `missing`.

The vault's durable decisions are recorded as ADRs:

- **ADR 0041**: the vault has one storage home at a time (device XOR vault), never an override resolver.
- **ADR 0042**: the vault keys to its owner in the persistence layer, never the guid; forgetting is a local detach. (This is the account-scoping decision below.)
- **ADR 0043**: the vault's encryption key comes from the passphrase, not the server; that key source is the only zero-knowledge lever.

## Vault Shape

The encrypted KV keeps the workspace shape local-first. Values are encrypted at the key-value layer, synchronized through the existing room machinery, and decrypted only when the caller has an unlocked vault keyring.

```ts
const secrets = createEncryptedLwwKeyValue({
  doc,
  keyring,
  namespace: "secrets",
});
```

The Whispering facade sits above that primitive. Today, with vault sync disabled, it preserves the existing device-only behavior. Reads fall back to `deviceConfig`; writes from the settings form still land there. Once the vault UI exists, the same consumers can distinguish `locked` from `missing` without calling provider SDKs with blank keys.

## Consumer Wiring

Wave 5 routes provider-key reads through the facade at the runtime edges: transcription, transformations, setup readiness, and the transformation editor warning. Non-secret settings stay on `deviceConfig`, including endpoints, model IDs, and local model paths.

The app layout now waits for `secrets.whenReady` before rendering child pages, so reactive reads cannot observe the unhydrated IndexedDB state. The readiness copy stays binary for use-now behavior: missing and locked are both not usable, but the remedy differs between "add your key" and "unlock your secret vault."

## Settings Write Surface (wave 6)

Wave 6 routes the last raw boundary, the settings key inputs in `ProviderConfigFields`, through `secrets.set`. The secret input renders every read state (`available`, `missing`, `locked`), so it is correct the moment the vault lifecycle lands, and a failed write surfaces loudly rather than dropping a key. Endpoints, base URLs, and model IDs are not secrets and stay on `deviceConfig`.

## Account scoping (the resolved design gate)

The vault syncs through the relay keyed to an owner, but Whispering has no login yet. The invariant: a vault must not lead its host app's identity. ADR 0042 resolves it, and the shipped code already follows it:

- **The guid is a stable per-app constant; the owner lives in the persistence layer, never the guid string.** No `guid:ownerId` concatenation.
- **With no account (Whispering today)**, the vault attaches via raw `attachIndexedDb` under one implicit, local, per-device owner, byte-for-byte the same persistence wiring as Whispering's own main workspace (`whispering.browser.ts`), which is not account-aware either. The vault matches its host rather than leading it: it invents no account and forces no auth model. This is the safe degenerate, local-only and per-device.
- **When auth lands**, the only change is the persistence attachment: swap `attachIndexedDb(vaultDoc)` for `attachLocalStorage(vaultDoc, { server, ownerId })`, which keys IndexedDB and the BroadcastChannel by `(server, ownerId, guid)`, mirroring the relay's `owners/<ownerId>/rooms/<guid>` partition, under the framework's dispose-on-sign-out / remount-on-sign-in lifecycle. The guid never moves.
- **`forget()` is a local detach, not a propagating destroy.** It is the lost-passphrase recovery path: it wipes this device's replica only. With no relay attached today, `vault.destroy()`'s CRDT deletes never leave the local IndexedDB, so it is already that detach; the code records the constraint that, under relay sync, `forget` must never propagate deletes, so forgetting on one device cannot destroy a sibling's still-unlockable vault. `disableSync` remains the graceful path that migrates values back; `forget` accepts their loss.

So the no-account path degrades correctly with no fix needed. The work this pass added is documentation: the three vault ADRs (the design was previously recorded only in another worktree's uncommitted ADRs, which the recent main merge renumbered out from under the code) and repointing every in-code ADR citation to them.

This change:

No encryption reaches a user in this PR, and that is deliberate. Whispering has no auth yet, so the vault runs as the local-only, per-device singleton described above and stays in its `device-only` home: with no provisioning UI it never leaves the `absent` state. Every provider key still lives in plaintext `localStorage`, exactly as before.

Because of that, no vault lifecycle controls ship: no enable-sync passphrase prompt, no unlock/lock/forget. A passphrase prompt now would provision a local-only vault while implying cross-device sync works. The `locked` state is therefore unreachable through the UI; it is wired everywhere so the account-aware wave that adds auth, owner-scoped persistence, and the lifecycle UI is the only piece left to make encryption real.

So this PR is the foundation and the seam: the encrypted-KV primitive, the passphrase keyring, a relay-sync proof, the facade, and every consumer (reads and writes) routed through it. The on-ramp lands with auth.

## Changelog

- Add an encrypted secret-vault foundation for Whispering provider credentials
- Route provider key reads and writes through the credential facade
- Show clearer missing-key guidance before cloud transcription and transformations call provider SDKs

---
<a href="https://app.blacksmith.sh/EpicenterHQ/codesmith/epicenter/pr/2056"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784219939&installation_id=128463665&pr_number=2056&repository=EpicenterHQ%2Fepicenter&return_to=https%3A%2F%2Fgithub.com%2FEpicenterHQ%2Fepicenter%2Fpull%2F2056&signature=63b86a51e59f026ea020986eed4fcc8903336a92e18e88694e4eccd5b6facb33"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


braden-w:
Superseded by #2220 (merged). That PR reshapes this same secret-vault foundation onto current main with a server-derived keyring instead of a passphrase vault (no unlock, no `locked` state), recorded in ADR-0074. The passphrase keyring is deferred as a premium seam. Closing this in favor of the merged work.

PR #2071, opened 2026-06-16
fix(whispering): teach the recording shortcut that actually fires

braden-w:
On a fresh desktop install, the home page and the record buttons showed an empty key hint where the recording shortcut should be. The shortcut that actually works there, push-to-talk on Fn, went untaught.

The reason for this shape is:

Manual recording has two start commands: push-to-talk (hold) and toggle (tap). Desktop binds push-to-talk (Fn) by default and ships the tap-toggle unbound; the browser ships the reverse. But every display resolved a single fixed command, `toggleManualRecording`:

- the action buttons called `shortcuts.label('toggleManualRecording')`
- the home page read `shortcuts.global.toggleManualRecording` (and the local toggle) straight out of `deviceConfig`/`settings`, the last surface still reaching into storage instead of the label seam it was told to use

So on desktop, where that command is unbound by default, the hint rendered empty even though Fn was live.

This change does the following:

- Add `getRecordingShortcutLabel(mode)`: resolves the first bound command in a per-mode preference list (`pushToTalk`, then `toggleManualRecording` for manual; `toggleVadRecording` for VAD) through the `#platform/shortcuts` label seam. One owner of "which gesture do we teach."
- Route both action buttons and the home page through it.
- Collapse the desktop home copy: the old `here` (local) / `anywhere` (global) split showed a local in-app shortcut that never registers on desktop. It is now one honest line per mode that hides when nothing is bound.

Stacked conceptually on the #2026 accessibility/shortcut rework and the #2069 label seam; this finishes routing the last raw-storage reader through the seam.

## Changelog

- Show the recording shortcut that is actually bound on each platform (push-to-talk on desktop), instead of an empty hint for the unbound toggle.

---
<a href="https://app.blacksmith.sh/EpicenterHQ/codesmith/epicenter/pr/2071"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784244909&installation_id=128463665&pr_number=2071&repository=EpicenterHQ%2Fepicenter&return_to=https%3A%2F%2Fgithub.com%2FEpicenterHQ%2Fepicenter%2Fpull%2F2071&signature=82c68e17a17a4c783af078130881bd74fcb993270c0e3a9a58c4218d19d15e5f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/whispering/src/lib/utils/recording-shortcut.ts
+++ b/apps/whispering/src/lib/utils/recording-shortcut.ts
@@ -1,0 +1,37 @@
+import { shortcuts } from '#platform/shortcuts';
+import type { Command } from '$lib/commands';
+
+/**
+ * Preference order for the shortcut that starts each recording mode: the first
+ * command with a binding live on this platform wins.
+ *
+ * Manual recording has two start commands. Desktop binds push-to-talk (Fn) by
+ * default and ships the tap-toggle unbound; the browser ships push-to-talk
+ * unbound and the toggle bound. So push-to-talk leads and the toggle backs it
+ * up, and whichever the user actually bound is the one we show. VAD has a single
+ * command, so its list has one entry.
+ */
+const RECORDING_SHORTCUT_PREFERENCE = {
+	manual: ['pushToTalk', 'toggleManualRecording'],
+	vad: ['toggleVadRecording'],
+} as const satisfies Record<string, readonly Command['id'][]>;
+
+export type RecordingShortcutMode = keyof typeof RECORDING_SHORTCUT_PREFERENCE;
+
+/**
+ * The display label for the shortcut that actually starts this recording mode on
+ * this platform, resolved through the `#platform/shortcuts` label seam.
+ *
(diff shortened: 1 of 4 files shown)


braden-w:
Superseded by #2074, which carries this commit unchanged and makes the recording-shortcut key a view over the DictationCapability (dims when macOS Accessibility is not granted). Closing in favor of the unified PR.

PR #2072, opened 2026-06-16
Rust owns the macOS dictation capability

braden-w:
The macOS Accessibility trust fact belongs in the Rust process that owns the `rdev` keyboard tap. The frontend could only guess at two OS-level facts it cannot observe directly: whether the tap is alive and whether a stale Accessibility grant is actually usable.

The old shape spread that truth across a third-party `AXIsProcessTrusted` probe, a frontend tri-state permission cache, a six-state listener supervisor that inferred liveness, and a focus/blur grant poll. The load-bearing bug was a stale post-update grant: `AXIsProcessTrusted` reported it as trusted, the broken tap read as `granted`, the home notice never appeared, and the user got generic "restart the app" guidance instead of the real fix, remove and re-add the grant.

## Rust-Owned Capability

A supervisor in `src-tauri/src/keyboard` owns the tap lifecycle:

- gates spawning on the live `AXIsProcessTrusted` check
- restarts a tap that dies under a held grant with capped backoff
- publishes one `DictationCapability`: `unknown`, `unsupported`, `untrusted`, `active`, or `broken`

The frontend seeds from `get_dictation_capability` and renders that value. It no longer probes the OS, infers liveness, or polls for grant changes. The one unavoidable poll, because macOS has no grant-flip event, now lives in Rust beside the tap and runs only while the capability is not `active`.

Because the supervisor sees a tap die while the grant still reads trusted, `broken` is distinguishable from `active` by construction. The home notice can now route a stale grant to the remove-and-re-add guide with its own copy.

## Removed Frontend Machinery

- `start_keyboard_listener` and `ListenerStart`
- `KeyboardListenerStoppedEvent`
- `permissions.svelte.ts`; the only consumers were accessibility, and microphone requests already go through `tauri.permissions.microphone`
- `global-listener.svelte.ts`, replaced by `dictation-capability.svelte.ts`
- the frontend `AXIsProcessTrusted` probe and bounded grant poll

The decision is recorded in [ADR-0010](https://github.com/EpicenterHQ/epicenter/blob/feat/whispering-dictation-capability/docs/adr/0010-rust-owns-the-macos-dictation-capability.md), settling the accessibility-grant and listener-liveness concern that [ADR-0008](docs/adr/0008-rdev-backs-the-desktop-global-trigger.md) deferred.

This branch predated a mainline model-folder Rust refactor that touched `lib.rs` and `bindings.gen.ts`, so a rebase and bindings regeneration against current `main` were expected before merge.

---
<a href="https://app.blacksmith.sh/EpicenterHQ/codesmith/epicenter/pr/2072"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784245078&installation_id=128463665&pr_number=2072&repository=EpicenterHQ%2Fepicenter&return_to=https%3A%2F%2Fgithub.com%2FEpicenterHQ%2Fepicenter%2Fpull%2F2072&signature=d0c765b3e319d674a31a84aab09de0ecfcd834f8769334e207f13fbe5b5d54f1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


braden-w:
Superseded by #2074, which carries these three commits unchanged and adds the capability-aware key dimming. Closing in favor of the unified PR.

PR #2083, opened 2026-06-18
perf(whispering): hand finalized recording PCM to transcription in-process

braden-w:
On desktop, the finalized audio for one recording is already a mono 16 kHz `Vec<f32>` sitting in Rust the moment you stop speaking, which is exactly what the local model and the Opus encoder consume. The live path used to throw that buffer away and rebuild it: write a WAV (with `fsync`), then read it back and Symphonia-decode it, all on the critical path between "stop" and "transcript". That round-trip recovers a buffer Rust held in RAM milliseconds earlier.

This routes the buffer in-process instead. The filesystem leaves the critical path; the WAV becomes a side effect.

**What `PcmHandoff` actually is (no overselling).** It is not "the AudioEngine" and it does not centralize routing: local-vs-cloud selection still lives in JS. Its real job is to be **the synchronization point that makes the off-path WAV write safe**. On the live path the stash is a guaranteed hit (`put` happens-before `stop_recording` returns, before JS can call `transcribe`), so the live consumer reads from memory and never races the concurrent persist. Without it, moving the write off the critical path would be a read-before-write bug. Honest cost: this **adds a path** (memory hit, else disk miss) per consumer rather than collapsing to one. That extra path pays for itself: it is the only way to take the blocking write off the critical path safely.

- `audio::PcmHandoff` is a Tauri-managed, single-slot store of the most recent finalized PCM, keyed by recording id. `stop_recording` stashes it; `transcribe_recording` and `encode_recording_for_upload` take it consume-once. A non-live miss falls back to decoding the WAV from disk, which is always correct, so the store self-evicts on the next stop (memory bounded to one recording). That eviction is a deliberate, documented relaxation of "owned until persist acks": the only live-path miss it can cause is a pathological rapid-fire (stop A, fully start+stop B, then A transcribes), which degrades to a graceful fall-back decode, never corruption.
- `artifact.rs` splits the pure handle (`artifact_handle`, deterministic `byte_length = 44 + n*4`) from the WAV write (`persist_artifact`). `stop_recording` builds the handle synchronously and spawns the write off the critical path. On failure Rust emits `recorder:persist-failed` and the frontend shows a transient "audio not saved" toast: the transcript still landed (from memory), the only loss is that recording's audio in history.

**What did not change:** audio bytes still never cross the JS/Tauri boundary (JS traffics the id-handle only), and `stop -> transcribe(id) -> deliver` is identical on the wire. The added `State<PcmHandoff>` params are erased from the generated bindings, so the only `bindings.gen.ts` delta is docstrings.

**Instrumentation (the actual deliverable).** A flag-gated `timing` module (`WHISPERING_TIMING=1`, log target `whispering::timing`, zero cost when off) puts spans on finalize, persist, decode-on-miss, and opus, plus handoff hit/miss markers and unified `model.load COLD/warm-reuse` + `model.inference` notes. These numbers are what decide whether PR 2 (prewarm) or PR 3 (Rust VAD) comes next.

**Measured (Apple Silicon, NVMe SSD)** via `cargo test file_roundtrip_overhead -- --nocapture`. This is the write+fsync + read+decode round-trip the PR removes from the live path, and **it is the only thing measured**:

| Clip | write + fsync | read + decode | total removed |
|---|---|---|---|
| 5 s  | ~9-15 ms  | ~4-12 ms  | **~14-27 ms** |
| 15 s | ~8-9 ms   | ~12 ms    | **~20-22 ms** |
| 60 s | ~27-29 ms | ~48-49 ms | **~75-79 ms** |

**What this PR does NOT claim.** End-to-end stop-to-delivery was not measured (an agent can't drive a mic + real model in CI), so the PR makes no claim about what fraction of perceived latency it removes. Two caveats this clean dev box cannot capture:

- The **fsync tail is the suspected real lever and is invisible here.** This NVMe never spiked; the variable 50-100+ ms (or worse) spikes live on contended disks: a cloud-synced recordings folder, Windows AV scanning the write, a spinning HDD. Async persist removes that blocking write from the path regardless of size. Whether it is a big lever depends on the user's disk, which the owed measurement must provoke.
- On a CPU-only box, inference dominates more, so the round-trip is proportionally smaller. Good thing the PR is cheap there.

The PR-2-vs-PR-3 call is therefore **not yet decidable from this data**; it waits on a real-hardware, warm-vs-cold, contended-vs-fast-disk measurement using the shipped `WHISPERING_TIMING` flag.

Part of the desktop-audio-pipeline direction in `apps/whispering/specs/20260617T170000-desktop-audio-pipeline-greenfield.md`. Supersedes and deletes `specs/transcription-latency-optimization.md`.


braden-w:
Closing in favor of a deliberate split (per review discussion): measure before optimizing.

This PR bundled instrumentation + the in-process handoff + async persist. Since the instrumentation is what justifies the optimization, shipping both at once merges an unproven change. So:

- **Instrumentation only** is now #2085 (one read path, one source of truth, zero new runtime surface). It ships first so we can measure stop->delivery and the fsync tail on real hardware.
- **The handoff + async-persist optimization** from this PR is preserved on `feat/whispering-pcm-handoff-optimization`, gated on #2085's measurement. It merges only if the numbers (round-trip share of stop->delivery, or an fsync tail on contended disks) earn it.

The branch `feat/whispering-inprocess-pcm-handoff` for this PR is now redundant with the parked branch and can be deleted.

PR #2127, opened 2026-06-20
refactor: split answerer resolution into owner ⊥ engine

braden-w:
Stacked on #2122 (now merged). Separates the two orthogonal questions a Zhongwen answerer resolves, which were previously tangled in one place, into the two axes named by ADR-0033/0038:

- **owner** (designation): *who writes the transcript?* An `'ephemeral'` agent (`this-device`) is owned by the open browser tab, which answers in-process and stops when it closes. A `'durable'` agent is an always-on resident daemon that answers over sync. Each peer decides designation where it naturally can: a daemon's observe loop hosts only conversations bound to its agent; a browser reads the bound agent's `owner` before mounting an answerer.
- **engine** (where tokens come from): a local key, the user's metered account. Resolved as a priority walk (`resolveEngine`) the same way on both peers, never a property of the owner.

This change:
- The agent catalog declares an `owner` kind (`'ephemeral' | 'durable'`), not a `runtime`: the routing fork is about who writes, not where the runtime lives.
- The ephemeral agent is renamed `epicenter-cloud` → `this-device`: the id names the *owner* (the device's tab, which is why the answer stops when you close it), not the engine (the cloud the tokens bill to). owner ⊥ engine.
- The engine walk is shared, leaving designation where each peer decides it (a daemon funnelling designation through the same function made it a tautology).
- The metered Epicenter engine is single-homed in `epicenterMeteredEngine(sessionFetch, baseURL)` on a new `@epicenter/zhongwen/engine` subpath: both peers (browser tab, signed-in daemon) build the same wire shape (AI-chat fetch wrap, `/api/ai/chat` route, `{ model, systemPrompts }`) from one place, kept out of the dep-free `zhongwen.ts` contract.

Unblocks the picker's live/offline presence decoration (which reads the `owner` fork) and the larger agentic tool-loop work, both as follow-ups.

---
[...]

PR #2128, opened 2026-06-20
feat(workspace,server): carry agent designation through presence

braden-w:
Slice 1 of daemon-liveness presence for the agent picker (ADR-0025).

The problem this addresses is:
A presence `Peer` announces `nodeId` + action manifest but **not which catalog agent it answers as**. The Zhongwen agent picker keys off `agent.id` (e.g. `zhongwen-home`), so with the peer list alone it can see *a* peer is online but can't tell that the resident daemon serving a given durable agent is live.

## Change
Carry an optional `agentId` on the presence wire as the **join key** the picker needs:

```
MountOptions.agentId (already first-class)
  → attachMountInfrastructure({ …, agentId })
    → openCollaboration({ …, agentId })
      → presence_publish frame { type, actions, agentId }
        → relay stores it on the Connection, echoes it in the peer snapshot
          → peers.list()[].agentId   ← picker join key
```

- `PeerSchema` / `PresencePublishFrameSchema` gain an optional `agentId` (TypeBox `Type.Optional`).
- The daemon's mount publishes its `agentId`; ordinary participants and content docs omit it.
- The relay stores it against the connection (like the action manifest) and mirrors it in `snapshotPeers`.

The reason for this shape is:
The node→agent binding is exactly as ephemeral as the connection. Putting it in a durable Y.Doc map would only go stale the moment a daemon dies (you'd still need presence to know it's live) and duplicate what the transport already owns. `agentId` is the join key only; the catalog stays the source of truth for an agent's label, model, and tools.

The scope is intentionally bounded:
Framework only; the Zhongwen mount already passes `agentId`, so this lights up end-to-end with no app change. The picker decoration that consumes it (reading `owner` from the catalog) is a follow-up slice on top of #2127.

---
<a href="https://app.blacksmith.sh/EpicenterHQ/codesmith/epicenter/pr/2128"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784567709&installation_id=128463665&pr_number=2128&repository=EpicenterHQ%2Fepicenter&return_to=https%3A%2F%2Fgithub.com%2FEpicenterHQ%2Fepicenter%2Fpull%2F2128&signature=f980134c2daf4ba7f4534ec1f3459f11031ff4f713d105399b02aad5edb90fba"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/server/src/room/backends/cloudflare/durable-object.test.ts
+++ b/packages/server/src/room/backends/cloudflare/durable-object.test.ts
@@ -266,6 +266,7 @@ type WirePeer = {
 	nodeId: string;
 	connectedAt: number;
 	actions: Record<string, unknown>;
+	agentId?: string;
 };
 type PresenceFrame = { type: 'presence'; peers: WirePeer[] };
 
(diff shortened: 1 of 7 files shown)


braden-w:
**Parked** per ADR-0041 + `specs/20260620T000000-vocab-answerer-collapse.md`. Presence decorates liveness but never gates binding (the conversation doc is a durable mailbox — an offline daemon still answers on wake), so it's not on the critical path for the answerer collapse. Revisit as a fast-follow after the daemon ships live (wave 2), reframed as the **dead-mailbox warning** ('you bound to your home daemon but nothing is home'). Please don't merge into the collapse stack.

PR #2140, opened 2026-06-21
chore: pin bun install linker to isolated

braden-w:
Add a root `bunfig.toml` pinning the bun install linker to `isolated` (pnpm-style).

The reason for this shape is:

bun's default linker varies by bun version, workspace setup, and the lockfile `configVersion`, so the `node_modules` layout was diverging across machines and could flip silently on a bun upgrade.

That already bit us. `onnxruntime-web` is a transitive dependency of `@ricky0123/vad-web`, not declared by whispering, so it is unreachable from the app root under an isolated tree. Whispering's vite config resolved it from the app root and threw `MODULE_NOT_FOUND` at config-eval time on an isolated machine while working fine on a hoisted one. The resolution path was fixed separately (resolve onnx through vad-web); this pins the layout so the hoisted-vs-isolated divergence stops happening across machines and CI.

`isolated` is the strict layout: a package can only import the dependencies it declares, so phantom dependencies fail loudly instead of riding on accidental hoisting.

A note for review:

This is repo-wide. The quality check on this PR exercises a full install and build under isolated. If another app was relying on a hoisted phantom dependency, it surfaces here rather than on `main`.

---
<a href="https://app.blacksmith.sh/EpicenterHQ/codesmith/epicenter/pr/2140"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784595449&installation_id=128463665&pr_number=2140&repository=EpicenterHQ%2Fepicenter&return_to=https%3A%2F%2Fgithub.com%2FEpicenterHQ%2Fepicenter%2Fpull%2F2140&signature=6dc4da03763ec9f72e01d1dd71e1bb7d2d2a870a105f78f9e3c104ec421cca85"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/whispering/src/lib/components/settings/LocalModelSelector.svelte
+++ b/apps/whispering/src/lib/components/settings/LocalModelSelector.svelte
@@ -266,7 +266,7 @@
 				</Item.Actions>
 			</Item.Root>
 		{:else if !value && customEntries.length === 0}
-			<Empty.Root class="py-8">
+			<Empty.Root>
 				<Empty.Media variant="icon">
 					<HardDriveDownload class="size-5" />
 				</Empty.Media>
--- a/apps/whispering/src/routes/(app)/+page.svelte
+++ b/apps/whispering/src/routes/(app)/+page.svelte
@@ -2,10 +2,14 @@
 	import { Button } from '@epicenter/ui/button';
 	import { confirmationDialog } from '@epicenter/ui/confirmation-dialog';
 	import { FileDropZone } from '@epicenter/ui/file-drop-zone';
+	import * as Item from '@epicenter/ui/item';
 	import * as Kbd from '@epicenter/ui/kbd';
 	import { Link } from '@epicenter/ui/link';
 	import * as SectionHeader from '@epicenter/ui/section-header';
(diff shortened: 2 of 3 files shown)


braden-w:
Splitting this PR. The bunfig pin moved to #2152, which is a clean one-commit PR containing only the `bunfig.toml` change.

The two whispering onboarding commits here (drop dead padding class, redesign onboarding as two-column) are superseded by #2141, which is built on top of them and replaces the two-column layout with single-column. The onboarding work ships through #2141 instead, so it should not merge via this PR.

PR #2143, opened 2026-06-21
refactor(whispering): collapse nullable-platform ceremony in settings

braden-w:
Whispering exposes Tauri-only capabilities through a single nullable object, `tauri` (`Tauri | null`, `null` on web). Most consumers use it imperatively (`if (!tauri) return`), which is fine. But the Launch-on-Startup toggle consumed it *reactively*: its TanStack query/mutation `.options` came from `tauri.autostart.*`, and because `createQuery`/`createMutation` cannot take a null options object, the settings page had to fabricate no-op fallback hooks and `tauri ? real : stub` ternaries purely for the web branch. It was the only place in the app doing this.

**Old shape** (`settings/+page.svelte`):

```svelte
const autostartQuery = createQuery(() =>
  tauri
    ? tauri.autostart.isEnabled.options
    : { queryKey: autostartKeys.isEnabled, queryFn: async () => false, enabled: false },
);
// ...two more fallback mutations like this, then an inline {#if tauri} switch block
```

**New shape** narrows `tauri` once at the boundary and prop-injects the non-null namespace into a colocated child that does zero null handling:

```svelte
{#if tauri}
  <AutostartSwitch autostart={tauri.autostart} />
{/if}
```

```svelte
<!-- AutostartSwitch.svelte -->
let { autostart }: { autostart: Tauri['autostart'] } = $props();
const isEnabledQuery = createQuery(() => autostart.isEnabled.options); // non-null
```

The null is handled exactly once, where TypeScript already narrows it; nothing downstream re-checks. This deletes the fallback ceremony and the now-orphaned `autostart-keys.ts`, which existed only so the web fallback could name the query keys without importing a Tauri-only module. The keys move next to their sole consumer in `tauri.tauri.ts`.

### Decision: keep the mega-object

It is tempting to read this as "dissolve `#platform/tauri` into per-capability seams." We deliberately did not. For capabilities with no meaningful web behavior (`tray`, `fs`, `keyboard`), the atomic `Tauri | null` is the right model: a browser no-op stub would pretend the capability exists and silently swallow calls that are actually bugs. Autostart was simply mis-homed, since it is the one capability consumed reactively, so it earned a colocated component rather than a new seam.

### Behavior tweaks

Two small truth-alignment changes ride along, both consistent with the OS login-item registry being the single source of truth (Whispering intentionally stores no shadow setting):

- Drop `initialData: false`. The switch is already disabled while the query is pending, so there is no wrong-OFF flash to guard against.
- Add `refetchOnWindowFocus`, so a login-item change made outside the app (System Settings, another tool) is reflected when the window regains focus.

### Related dead-code removal

The whispercpp, parakeet, and moonshine branches of `TranscriptionRuntimeConfig` each wrapped their `LocalModelSelector` in `{#if tauri}`. An earlier branch in the same if-chain (`{#if isSelectedServiceUnavailable && selectedTranscriptionProvider}`) already matches the web case (a local service with `!tauri` is unavailable) and short-circuits, so the local-engine branches are only reachable when `tauri` is truthy. The guards were unreachable and are removed.

### Further reading

This is a concrete application of the prop-drilling lesson in [`tauri` is both the namespace and the platform check](https://github.com/EpicenterHQ/epicenter/blob/main/docs/articles/20260526T012526-tauri-is-both-the-namespace-and-the-platform-check.md), specifically its "Pushing the narrowing further: prop drilling" section: narrow `tauri` once at the boundary, then pass the non-null namespace down so nothing below re-checks. `AutostartSwitch` is now a second worked example of that move alongside `ShortcutTable`. See also [Two files, one import](https://github.com/EpicenterHQ/epicenter/blob/main/docs/articles/20260525T234034-two-files-one-import-build-time-platform-injection.md) on why runtime platform ternaries rot: "by the tenth service that does the same dance, you stop reading the conditional and start treating it as noise."

### DIFF
--- a/apps/whispering/src/lib/components/settings/TranscriptionRuntimeConfig.svelte
+++ b/apps/whispering/src/lib/components/settings/TranscriptionRuntimeConfig.svelte
@@ -307,99 +307,93 @@
 		</div>
 	{:else if settings.get('transcription.service') === 'whispercpp'}
 		<div class="space-y-4">
-			{#if tauri}
-				<LocalModelSelector
-					models={WHISPER_MODELS}
-					title="Whisper Model"
-					description="Download a pre-built model or add your own to the models folder. Models run locally for private, offline transcription."
-					bind:value={() => deviceConfig.get('transcription.whispercpp.model'),
-						(v) => deviceConfig.set('transcription.whispercpp.model', v)}
-				>
-					{#snippet footer()}
-						<Field.Description>
-							Pre-built models are downloaded from
-							<Link
-								href="https://huggingface.co/ggerganov/whisper.cpp"
-								target="_blank"
-								rel="noopener noreferrer"
-							>
-								Hugging Face
-							</Link>
-							into the models folder. Quantized models (q5_0, q8_0) offer
-							smaller sizes with minimal quality loss.
-						</Field.Description>
-					{/snippet}
-				</LocalModelSelector>
-			{/if}
+			<LocalModelSelector
+				models={WHISPER_MODELS}
+				title="Whisper Model"
+				description="Download a pre-built model or add your own to the models folder. Models run locally for private, offline transcription."
+				bind:value={() => deviceConfig.get('transcription.whispercpp.model'),
+					(v) => deviceConfig.set('transcription.whispercpp.model', v)}
+			>
+				{#snippet footer()}
+					<Field.Description>
+						Pre-built models are downloaded from
+						<Link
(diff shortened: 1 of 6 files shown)


braden-w:
Retrospective note: this is a concrete example of “let the guard carry the capability.” The important part is that `tauri` is not just a runtime flag. Once `{#if tauri}` narrows it, it becomes the typed capability object, so `AutostartSwitch` can receive the already-proven autostart namespace instead of rebuilding a nullable/web-fallback world downstream.

PR #2201, opened 2026-06-26
refactor: collapse the per-app instance auth into one shared resolver and settings UI

braden-w:
Every browser and extension app hand-repeated the same three things to answer "which Epicenter server do I talk to, and how do I authenticate": a per-app `instance.ts` (read/write/clear over storage), a byte-identical `InstanceSettingsModal`, and the `instance.token ? createInstanceTokenAuth(...) : createOAuthAppAuth(...)` ternary. The node/CLI side already collapsed that fork into one choke point (`createMachineAuthClient`); the browser and extension side never did. This collapses it.

## One resolver, one setting, one UI

A persisted `Instance` (`{ baseURL, token? }`) now resolves through a single client-side choke point, mirroring the node resolver:

```ts
// before, hand-repeated in every app:
const auth = instance.token
  ? createInstanceTokenAuth({ baseURL: instance.baseURL, token: instance.token })
  : createOAuthAppAuth({ baseURL: instance.baseURL, clientId, persistedAuthStorage, launcher });

// after, the same call in every app:
const auth = createAppAuthClient(instanceSetting.read(), { clientId, persistedAuthStorage, launcher });
```

The storage difference (synchronous `localStorage` for the SvelteKit apps, async `chrome.storage` for the extension) hides behind one injected handle. A per-app `instance.ts` is now the whole file:

```ts
export const instanceSetting = createInstanceSetting({
  storageKey: 'fuji.instance',
  defaultBaseURL: APP_URLS.API,
  storage: globalThis.localStorage,
});
```

The hand-copied modal and signed-out gate become `@epicenter/app-shell/instance-settings` components that take an `appName` plus the injected setting handle. fuji, opensidian, vocab, honeycrisp, and the tab-manager extension all route through them, and the reactive `createOAuthAppAuth` / `createInstanceTokenAuth` wrappers in `@epicenter/svelte` collapse into one reactive `createAppAuthClient`.

## OAuth is hosted-only now (ADR-0071)

The asymmetric refusal that makes the collapse clean: OAuth runs only against the hosted star, so a self-hosted instance requires a token (ADR-0070's first-boot bearer). This deletes per-instance OAuth issuer and resource derivation and the launcher-by-base-URL factory; every launcher is built once from the app's hosted constants.

- Before: a no-token instance could run OAuth against any base URL, so a self-hoster had to register a Google OAuth app for their own box.
- After: a no-token instance is the hosted default; a self-hosted base URL carries the token its box minted, and the settings modal requires that token.

[...]

PR #2207, opened 2026-06-26
feat(self-host): mint a first-boot bearer so a solo box needs no OAuth app

braden-w:
A self-hoster should be able to reach their own box without registering a Google OAuth app against it. Until now they could not: the only wired sign-in was Google OAuth, so running your own machine meant taking a cloud dependency on Google's console, the exact wart ADR-0070 and ADR-0071 were written to remove. This is the server half of that fix (privacy-spec wave 3): a solo self-host star mints and persists its own single-user bearer at first boot and names the file for the operator to paste into the client instance setting.

The implementation works like this:

On first boot a solo box mints a bearer, writes it `0600` to `<DATA_DIR>/instance-token`, and names the file. It never echoes the token to stdout: journald / docker logs are a retained, shipped, worse at-rest location than the `0600` file.

```
apps/self-host (Bun) listening on http://localhost:8787 (rooms in /var/lib/epicenter)
solo mode. Minted a new instance token, saved 0600 to /var/lib/epicenter/instance-token.
Read it once and paste it into the client instance setting:
  cat /var/lib/epicenter/instance-token
```

The operator `cat`s it once and pastes it into the client's instance setting (`{ baseURL, token }`, ADR-0071). The resolution order is the canonical zero-config-but-secure pattern (code-server, n8n, Jupyter, Vault): an `INSTANCE_TOKEN` env wins verbatim for 12-factor / container-secret injection, else the persisted file is reused (a later boot prints "loaded", not "minted"), else a freshly minted `randomBytes(32).base64url`. The token and the room data resolve to one shared `DATA_DIR` (created on first boot), so persisting that volume keeps both and an ephemeral one loses both, never the credential alone. Every request then arrives as `Authorization: Bearer <token>`; `createInstanceTokenResolver` is the injected `ResolveUser` (a constant-time SHA-256 compare) that turns it into the box's single owner under `personal()`, feeding the one total gate exactly like the OAuth resolver (ADR-0072).

## The box's mode is an explicit launch choice

`EPICENTER_MODE` (`solo` | `shared`, default `solo`) picks the box's shape. It is an explicit operator declaration, **not** derived from which OAuth secrets happen to be set (the first cut did the latter, and this PR replaces it). The mode IS the data partition (`personal()` keys `owners/self-host`, `shared` keys `owners/shared`), so sniffing it from mutable credentials would let adding or rotating a secret silently re-partition a populated box and orphan its data. Instead the configured providers must agree with the declared mode, checked loudly at boot:

```
EPICENTER_MODE=shared  ->  shared wiki: OAuth + shared({ admit }) + ALLOWED_MEMBER_EMAILS  (>= 1 provider required)
unset / EPICENTER_MODE=solo  ->  solo box: first-boot bearer + personal()
```

A half-configured provider pair, `shared` with no provider, and `solo` carrying OAuth creds or an allowlist each fail boot with a message naming the fix. This is provider-agnostic: a GitHub-only, Google-only, and multi-provider wiki are all `EPICENTER_MODE=shared`. The solo-only knobs (`INSTANCE_TOKEN`, a leftover token file) are inert in shared mode, so they warn rather than fail. Because mode is an operator declaration revalidated against the creds every boot, it is not the stored discriminator ADR-0066/0071 reject.

## Contract change: OAuth providers are register-when-present

`ServerBindings` now marks `GOOGLE_CLIENT_ID` / `GOOGLE_CLIENT_SECRET` optional, joining GitHub. `createAuth` registers each provider only when configured, so a solo box builds a Better Auth instance with no social providers. `configuredSocialProviders(env)` stays the single set `createAuth` registers; it no longer selects the partition, and a new `incompleteSocialProviders(env)` rejects a half-set pair at boot. The hosted star is never provider-less, so `apps/api` re-requires Google in its own boot validation:

```ts
// apps/api/server.ts: keep the hosted guarantee even though the contract loosened
const env = BunHostBindings.merge({
  GOOGLE_CLIENT_ID: 'string',
  GOOGLE_CLIENT_SECRET: 'string',
})(process.env);
```

The Worker's generated `Cloudflare.Env` keeps Google required, so its `satisfies ServerBindings` is unaffected. No endpoint or WebSocket change was needed: `/api/session` already returns the right shape once `resolveUser` is injected, and the client (PR #2201) already sends the bearer.

The durable decisions are recorded in ADR-0072, which pins the default ADR-0070 left to the spec and sharpens ADR-0071.

## Changelog

- Self-hosters can now run a solo box with no Google OAuth app: the server mints and persists a single-user bearer at first boot (override with `INSTANCE_TOKEN`) and names the `0600` file for the operator to paste into the client, never echoing the secret to the logs. The box runs solo or shared-wiki by an explicit `EPICENTER_MODE` launch choice (default `solo`), validated against the configured OAuth providers at boot, rather than inferred from which secrets are set.

---
<a href="https://app.blacksmith.sh/EpicenterHQ/codesmith/epicenter/pr/2207"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785091004&installation_id=128463665&pr_number=2207&repository=EpicenterHQ%2Fepicenter&return_to=https%3A%2F%2Fgithub.com%2FEpicenterHQ%2Fepicenter%2Fpull%2F2207&signature=bdff703f8990457118ea56bfbc64a3683f6f4c7d553f82002a95ebf75b04fc68"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


braden-w:
Closing without merging: this approach is superseded before it ships.

The single-partition-instance decision on `feat/self-host-single-partition-instance` replaces it. That branch records this PR's design as ADR-0072 (Superseded) and its successor as ADR-0073: self-host is a single-partition instance behind one operator-supplied bearer, multi-tenancy is Cloud-only.

The successor deletes roughly 80% of this PR's surface. Gone: `EPICENTER_MODE` and the boot-agreement guard, first-boot minting, the 0600 token file, the TOCTOU hardening, the Bun-only mint constraint, and OAuth/sessions/allowlist on self-host. In their place: ONE partition pinned to a constant owner id (`owners/instance`), authenticated by ONE operator-supplied static `INSTANCE_TOKEN` (constant-time compared), with a fail-closed entropy gate holding the 256-bit floor. "Solo" vs "shared" stops being config, it is just how many people you hand the token to. The instance then runs identically on Bun and Cloudflare because the operator supplies the secret.

What carries forward: the constant-time bearer compare, and the finding that a bearer is the right primitive.

Note for the follow-up PR: main has already taken ADR numbers 0072 (local-books standalone CLI) and 0073 (MCP-native tools), so the single-partition branch must renumber its ADRs from 0072/0073 to 0074/0075 before it lands. Not deleting this branch.

PR #2243, opened 2026-07-01
feat(recorder): share browser recording and VAD

braden-w:
Whispering had the reusable half of recording trapped inside one app: browser stream acquisition, the MediaRecorder path, Silero VAD, and level-meter smoothing. Vocab needs that same microphone capability for dictation, so this extracts the portable runtime into `@epicenter/recorder` and proves the boundary with Vocab as a second consumer.

The package owns browser recording and VAD, but not app policy. It reads no settings store, writes no tables, runs no transcription, and renders no UI:

```ts
import {
  createBrowserRecorder,
  createVadRecorder,
  enumerateDevices,
  foldMicLevel,
} from "@epicenter/recorder";
```

Whispering keeps the native Tauri/CPAL recorder behind its existing platform seam. The browser branch now instantiates `createBrowserRecorder()`, while VAD uses the shared browser implementation in both web and the Tauri webview. The shared package also exposes `@epicenter/recorder/vad-assets`, a Node-side helper for Vite configs to copy the Silero and onnxruntime files into `/vad/` so runtime asset loading is explicit in each app.

Vocab now uses `createVadRecorder()` for continuous dictation: tap the mic to open a session, each pause-delimited phrase transcribes through the existing connection registry, and the resulting text lands in the chat input. The old in-app push-to-talk recorder is deleted instead of preserved as a second microphone owner.

The main trade-off is that the shared VAD core stays callback-based rather than shipping a Svelte wrapper. That keeps the package portable and lets each app own its reactive surface; a `/svelte` wrapper can earn its way in if a third consumer repeats the same state shape.

## Changelog
- Add microphone dictation to Vocab
- Share Whispering browser recording and voice activity detection through `@epicenter/recorder`


braden-w:
Closing this without merging.

The extraction's stated premise was "Vocab needs the same dictation capability." Vocab has since shipped its own recorder (`apps/vocab/src/lib/state/recorder.svelte.ts`, merged 2026-06-26), and its own code documents the opposite decision: it deliberately does not depend on a shared recorder package, on the grounds that Whispering's recorder is an app, not a capability, and that extraction should wait for a real second consumer of the same shape.

Reading both recorders side by side, the actual overlap is a six-line idiom (`getUserMedia`, `new MediaRecorder`, collect chunks, build a `Blob`), and the two consumers fill in every other knob differently:

- Whispering: WHISPER mic constraints, device enumerate/exact/fallback with an acquisition outcome, mime-type detection for Firefox/Safari quirks, level metering, Silero VAD, a `RecordingSession` lifecycle, native CPAL, ~10 error variants.
- Vocab: `{ audio: true }`, no device picker, no forced mime type (deliberately), no levels, no VAD, plain start/stop, 3 error variants.

There is no non-trivial shared surface, so a shared package would exist to serve sharing that is not happening: a single-consumer package plus an interaction surface (`mimeType?`, `constraints?`, `withLevels?` knobs) where two unlike needs negotiate through a shape neither owns. The cleaner shape is two cohesive in-app recorders with zero shared boundary.

Nothing is lost by closing. Every non-trivial piece (mime detection, device fallback, the `foldMicLevel` curve, VAD) already lives in Whispering on `main`; this PR only moved it. The spec added here dies with the branch.

Revisit trigger: extract when a third consumer wants one of these exact flavors (two real consumers of one shape), not merely when a third thing happens to record audio. Vocab's own recorder comment already commits to that rule.

PR #2364, opened 2026-07-04
feat(whispering): access-sectioned transcription setup catalog

braden-w:
Whispering now has two different transcription jobs: setup and switching. The recorder popover from #2353 owns the active route because that is where recording starts. This PR makes the settings page own the other half: getting every route ready without hiding setup behind the currently selected provider.

The transcription settings block is now an access-sectioned setup catalog, rendered in ACCESS_GROUPS order:

- On your device: Tauri-only local GGUF setup through the existing LocalModelSelector.
- Epicenter: connected-account state plus inline sign-in. Reauth-required becomes Reconnect, and the sign-in action is locked while recording because the auth flow reloads the page.
- Provider API: one setup card per keyed provider, with API-key fields and the provider model picker together.
- Custom server: the existing Speaches endpoint and model ID setup.

The important boundary is unchanged: this catalog never writes transcription.service. Adding a key, downloading a GGUF, signing in, or entering a Speaches endpoint makes a route ready; the recorder popover still chooses which ready route is active. The settings page reflects that active family or provider with a read-only Active badge so the user has orientation without creating a second switcher.

A small hardening pass is included because the Epicenter section observes TanStack mutation error state. The auth methods return Wellcrafted Results, so the account actions now use wellcrafted/query mutationOptions instead of returning Result objects directly or hand-throwing inside createMutation. That keeps Result unwrapping in the shared query adapter and lets failed sign-in or sign-out land in the normal mutation error channel.

The old TranscriptionServiceSelect dropdown is deleted, the now-dead grouped provider helpers are removed, and the spent reconciliation spec is deleted now that both phases have landed.

## Changelog

- Replaces Whispering's transcription provider dropdown in settings with access-sectioned setup panels.
- Adds inline Epicenter sign-in from the transcription setup section.
- Keeps active transcription switching in the recorder popover.


braden-w:
Reconciled the duplicate Phase 2 branch into this PR. This branch remains the base because it keeps the access-family presentation owner in ACCESS_GROUPS, preserves the read-only Active orientation, and deletes the spent spec. The follow-up commit ports the useful piece from #2368: richer Epicenter sign-in handling in the setup catalog, including reauth-required to Reconnect, recording-active lockout, and proper Result unwrapping so auth failures surface in the mutation error state.

PR #2367, opened 2026-07-04
fix(whispering): describe loopback-bonded session transcription as on-device

braden-w:
A self-host Epicenter instance bonded at loopback is still this machine. Before this PR, Whispering treated every `session` transcription as remote and showed "Audio is sent to Epicenter," even when `auth.baseURL` was `http://localhost:...` or `127.0.0.1`.

The resolver now receives the bonded deployment origin and classifies `session` with the same loopback rule already used for endpoint providers. `localhost`, `127.0.0.1`, and `[::1]` read as on-device; hosted URLs, LAN IPs, and other remote origins still read as sent to Epicenter.

The important cleanup is that locality no longer depends on sign-in state. Both production callers pass `auth.baseURL` directly, and `sessionBaseUrl` is required in the pure resolver API so a future caller cannot accidentally omit the origin and recreate the old remote-only behavior. Sign-in remains a usability question elsewhere; this code only answers where the audio goes.

### DIFF
--- a/apps/whispering/src/lib/components/settings/TranscriptionRuntimeConfig.svelte
+++ b/apps/whispering/src/lib/components/settings/TranscriptionRuntimeConfig.svelte
@@ -41,10 +41,9 @@
 		describeTranscriptionDestinationFromConfig({
 			service: settings.get('transcription.service'),
 			getDeviceConfig: deviceConfig.get,
-			// Session locality follows the bonded deployment: a signed-in session at a
-			// loopback base URL is this machine, so audio stays on-device.
-			sessionBaseUrl:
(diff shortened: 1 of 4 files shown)


braden-w:
Heads-up on merge ordering: #2364 (transcription setup catalog) rewrites `TranscriptionRuntimeConfig.svelte`. The core fix here (resolver + tests + `run-polish.ts`) lives in files #2364 does not touch, so it merges independently. The only overlap is the one `sessionBaseUrl:` line added to the `destination` derived block in that component; whichever PR merges second should carry that line into the surviving version. #2364's component already imports `auth`, so it's a trivial one-liner.

Issue #2432, opened 2026-07-09
The build is broken on Windows


historydev:
Bad commit: [https://github.com/EpicenterHQ/epicenter/commit/949a8f39](https://github.com/EpicenterHQ/epicenter/commit/949a8f39)
Good commit: https://github.com/EpicenterHQ/epicenter/commit/eca8caed

Path:
<details>
G:\Desktop\projects\epicenter\apps\whispering\src-tauri\target\debug\build\transcribe-cpp-sys-27236ecd8590fb0a\out\build\ggml\src\ggml-vulkan\vulkan-shaders-gen-prefix\src\vulkan-shaders-gen-stamp\vulkan-shaders-gen-configure-err.log
</details>

Error root:
<details>

```console
G:\Desktop\projects\epicenter\apps\whispering\src-tauri\target\debug\build\transcribe-cpp-sys-27236ecd8590fb0a\out\build\ggml\src\ggml-vulkan\vulkan-shaders-gen-prefix\src\vulkan-shaders-gen-build\CMakeFiles\CMakeScratch\TryCompile-iadk2e\testCCompiler.c(1,1): error C1083: Cannot open compiler generated file: '': Invalid argument [G:\Desktop\projects\epicenter\apps\whispering\src-tauri\target\debug\build\transcribe-cpp-sys-27236ecd8590fb0a\out\build\ggml\src\ggml-vulkan\vulkan-shaders-gen-prefix\src\vulkan-shaders-gen-build\CMakeFiles\CMakeScratch\TryCompile-iadk2e\cmTC_fdee1.vcxproj]
```

</details>

<details>
```console

S G:\Desktop\projects\epicenter\apps\whispering> bun --bun run dev
$ bun run dev:desktop
$ bun run scripts/launch-dev.ts
$ tauri dev --config src-tauri/tauri.dev.conf.json
     Running BeforeDevCommand (`bun run dev:web`)
$ vite dev

  VITE v7.3.5  ready in 1252 ms

  ➜  Local:   http://localhost:1420/
[vite-plugin-static-copy] Collected 4 items.
     Running DevCommand (`cargo  run --no-default-features --color always --`)
        Info Watching G:\Desktop\projects\epicenter\apps\whispering\src-tauri for changes...
   Compiling webview2-com-sys v0.38.0
   Compiling transcribe-cpp-sys v0.1.0
   Compiling tao v0.35.3
   Compiling tauri-winrt-notification v0.7.2
[...]

Issue #2421, opened 2026-07-08
Unable to build the desktop application on Windows


historydev:
.

PR #2498, opened 2026-09-07
fix(epicenter): make the Tauri host compile and its tests run on Windows

nqwrc:
`apps/epicenter`'s Tauri host does not compile on Windows, and once it does, its test executables cannot start. Four failures that look unrelated share a plain explanation: nothing exercises this crate on Windows yet. None of the eight workflows in `.github/workflows` asks for a Windows or a macOS runner, and `auto.release.yml` sits behind `if: false` until v8.0.0 is tagged, so the platform's requirements have not had an occasion to be met.

The first failure is `tauri-build` panicking before any first-party code compiles: only `icon.icns` and `icon.png` are tracked, and generating the Windows resource file needs `icons/icon.ico`. I generated it from the tracked `icon.png` at the seven standard sizes, so it is the same artwork rather than new branding.

The second is `RunEvent::Reopen`, which is the `NSApplicationDelegate`'s `applicationShouldHandleReopen`. Tauri declares the variant `#[cfg(target_os = "macos")]`, so matching it unconditionally in `src/lib.rs` is a hard `E0599` on every other target. It now carries the same cfg.

The third only bites debug builds. `audiopus_sys` compiles Opus through the `cmake` crate, which derives `CMAKE_BUILD_TYPE` from the package's `OPT_LEVEL` and `DEBUG`, so a dev build links the debug CRT while rustc always links the release one and the link dies on unresolved `__imp__CrtDbgReportW`. A dev-profile override builds that one C dependency optimized, which is the behaviour anyone would want regardless: nobody steps through Opus in a debugger.

The fourth is why no Windows test run has ever reported anything at all. `tauri-plugin-dialog` imports `TaskDialogIndirect`, which exists only in comctl32 v6. The app binary declares that side-by-side dependency through the manifest `tauri-build` embeds, but cargo's test executables link the same crate graph without one, so the loader binds the ancient v5.82 and every test executable dies at startup with `STATUS_ENTRYPOINT_NOT_FOUND` before a single test runs. `build.rs` now emits `/MANIFEST` plus the Common-Controls v6 dependency for every linked executable. The external `.exe.manifest` the linker writes is honoured because rustc embeds none, and the app binary is unaffected, since an embedded manifest wins over the external file.

With all four in place the crate checks clean and the unit-test suite executes end to end. What still fails there is confined to `app_data` and `recorder` — platform-root resolution and blob staging, which are genuine Windows behaviour gaps rather than build problems. They deserve their own change; this one is only about getting the crate to compile and its tests to start.

One thing worth knowing before you try it: building here also wants a shell carrying the MSVC environment, the Ninja generator (the Visual Studio generator cannot build `ggml-vulkan`'s nested `vulkan-shaders-gen` sub-build), a short `CARGO_TARGET_DIR` (the shader object paths run past `CMAKE_OBJECT_PATH_MAX` under the repo's own target dir), and `CMAKE_POLICY_VERSION_MINIMUM=3.5` for the Opus that `audiopus_sys` vendors. That is environment setup rather than a code change, so it is not in here, but I have a small batch file that sets all four and can send it separately if you would like it in the tree.

## Changelog

- Fix the desktop host failing to compile on Windows

---
<a href="https://app.blacksmith.sh/EpicenterHQ/codesmith/epicenter/pr/2498"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791415697&installation_model_id=20236&pr_number=2498&repository=EpicenterHQ%2Fepicenter&return_to=https%3A%2F%2Fgithub.com%2FEpicenterHQ%2Fepicenter%2Fpull%2F2498&signature=6595eb68f359675982dce55acc70b594285749872bd1d474dfaf66cb5c4c7aae"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>


nqwrc:
One piece of context I should have put in the body: #2432 ('The build is broken on Windows') was closed as completed, but it was closed by its reporter with an environment workaround - moving the cargo target directory near the disk root, because the vendored ggml-vulkan shader paths run past the 260-character limit. That is the same `CARGO_TARGET_DIR` prerequisite I mention at the end of the description, and no code landed for it. None of the four failures in this PR were covered by it, and #2421 has an empty body, so neither closed issue overlaps with what this changes.

nqwrc:
A correction to the numbers at the end of the description, and a pointer to where the rest went.

I wrote "109 passed, 9 failed". The right figure is **110 passed, 8 failed**: my worktree had never had `bun install` run in it, and `app_data`'s equality test shells out to `bun -e` to ask the TypeScript authority for the root, so it was failing on a missing module rather than on anything about Windows. With the dependencies installed it passes. Calling all nine "genuine Windows behaviour gaps" was therefore wrong about one of them, and the count was off by one.

Of the 8 that really do fail, 7 are the recorder chain and one is unrelated to this PR. Both are addressed separately in #2502, which takes the same tree to 118 passed, 1 failed — the one being a pre-existing disagreement between Rust's `Path::is_absolute` and Node's `isAbsolute` about bare rooted paths on Windows, described there and deliberately not fixed.

